Issue #56, opened 2015-05-24
A jsonb field problem


leedstyh:
Sorry for my bad English first!

I have a `struct` like this:

```
type User struct {
    //...
    Settings *UserJsonb // it's a jsonb field in pg
}

type UserJsonb struct {
	Timezone    string
	Stockparams []float64
}

func (m *UserJsonb) Scan(value interface{}) error {
	return json.Unmarshal(value.([]byte), m)
}

func (m UserJsonb) Value() (driver.Value, error) {
	b, err := json.Marshal(m)
	if err != nil {
		return nil, err
	}
	return string(b), nil
}

```

And it works well at first (both v2 and v3). My pg version is v9.4.

But after I updated `gopkg.in/pg.v3`(go get -u) and postgresql (`apt-get upgrade`), my app got an error message: `json:Unmarshal(nil *mods.UserJsonb)`.

Then I changed `Settings *UserJsonb` to `Settings UserJsonb`, it worked again!

So is it a bug? Or it just works like this?

vmihailenco:
That looks like a bug. Fixing.

vmihailenco:
Please try latest version. The issue should be fixed. Thanks for the report.

Also note that v3 does not require `Scan` and `Value` methods any more. I.e. following should work too:

```
package main

import (
	"fmt"

	"gopkg.in/pg.v3"
)

func pgdb() *pg.DB {
	return pg.Connect(&pg.Options{
		User:     "postgres",
		Database: "test",
	})
}

func main() {
	db := pgdb()

	var user User
	_, err := db.QueryOne(&user, `SELECT '{"timezone":"mytimezone"}'::jsonb AS settings`)
	if err != nil {
		panic(err)
	}
	fmt.Printf("%#v\n", user.Settings)
}

type User struct {
	Settings *UserJsonb // it's a jsonb field in pg
}

type UserJsonb struct {
	Timezone    string
	Stockparams []float64
}
```

leedstyh:
Thanks for the quick fixing.

BUT as to `v3 does not require Scan and Value`, I have new issue:

Read from db works well, but update db will panic.

Error message is `panic: pg: unsupported src type: <main.UserJsonb Value>`

```
package main

import (
    "fmt"
    "time"
    "strings"
    // "database/sql/driver"
    // "encoding/json"

    "gopkg.in/pg.v3"
)

func pgdb() *pg.DB {
    return pg.Connect(&pg.Options{
        User:     "pg",
    })
}

[...]

PR #59, opened 2015-05-25
append: format struct as json by default.

vmihailenco:


vmihailenco:
Fixes https://github.com/go-pg/pg/issues/56.

PR #97, opened 2016-01-05
CopyFrom now can be used in transaction

Babazka:
Previously CopyFrom always executed in separate connection.

### DIFF
--- a/db.go
+++ b/db.go
@@ -313,64 +313,11 @@ func (db *DB) CopyFrom(r io.Reader, q string, args ...interface{}) (Result, erro
 	if err != nil {
 		return nil, err
 	}
-
-	if err := writeQueryMsg(cn.buf, q, args...); err != nil {
-		db.pool.Put(cn)
-		return nil, err
-	}
-
-	if err := cn.FlushWrite(); err != nil {
-		db.freeConn(cn, err)
-		return nil, err
-	}
-
-	if err := readCopyInResponse(cn); err != nil {
-		db.freeConn(cn, err)
-		return nil, err
-	}
-
-	ready := make(chan struct{})
-	var res Result
-	go func() {
-		res, err = readReadyForQuery(cn)
-		close(ready)
-	}()
-
-	for {
-		select {
-		case <-ready:
-			break
-		default:
-		}
-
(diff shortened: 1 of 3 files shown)


Babazka:
@vmihailenco can you take a look, please?

vmihailenco:
Now `COPY FROM` is always executed in transaction, which does not look right. Do I miss anything?

Babazka:
Yes, `DB.CopyFrom` now executes in a single-statement transaction. It is functionally identical to previous behaviour, that is, to executing in autocommitted transactionless connection. And we now also get `Tx.CopyFrom` which runs within user-provided transaction.

vmihailenco:
I understand the thing about auto-commit, but then we can as well wrap every other query in transaction too and claim that it is the same.

Can you please extract `copyFrom(cn *conn, r io.Reader, q string, args ...interface{})` and call it from `db.CopyFrom` and `tx.CopyFrom`?

Babazka:
Done. How does it look now?

db.go
```diff
@@ -488,3 +435,58 @@ func assertOneAffected(res Result, coll *singleRecordCollection) (Result, error)
...
+
+		if err := cn.FlushWrite(); err != nil {
+			return nil, err
+		}
+	}
+
+	select {
+	case <-ready:
+	default:
+	}
```
vmihailenco: I wish I remember why this is needed :)

db.go
```diff
@@ -313,66 +313,13 @@ func (db *DB) CopyFrom(r io.Reader, q string, args ...interface{}) (Result, erro
...
-	<-ready
+	res, err := copyFrom(cn, r, q, args...)
 	if err != nil {
 		db.freeConn(cn, err)
 		return nil, err
 	}
-
 	db.pool.Put(cn)
-	return res, nil
+	return res, err
```
vmihailenco: `return res, nil`?

Babazka:
> return res, nil?

Fixed.

I cannot comment on ``select`` magic, copied the code as-is :)

vmihailenco:
Thanks for PR. Tagged v3.5.4.

PR #112, opened 2016-03-09
Rewrite examples.

anmic:


example_test.go
```diff
@@ -203,66 +203,159 @@ func ExampleDB_WithTimeout() {
 }
 
 func ExampleDB_Create() {
```
vmihailenco: ExampleDB_Create_books

example_test.go
```diff
@@ -203,66 +203,159 @@ func ExampleDB_WithTimeout() {
...
-	// title 1
-	// title 2
-	// title 3
+	book1 := Book{
+		Title:    "title 1",
```
vmihailenco: book 1

PR #173, opened 2016-04-07
User can choice quote of params mapping;

liut:
In some stage, like follow:
```go
TableName string `sql:"star_?SHARD_suffix" json:"-"`
```

will raise db err: `ERROR #42P01 relation "star_'earth'" does not exist`.

and same error with multi-schema:

```go
TableName string `sql:"?shard.users" json:"-"`
```
err: `ERROR #42P01 relation "'shard_9'.users" does not exist`.

### DIFF
--- a/orm/format.go
+++ b/orm/format.go
@@ -60,7 +60,7 @@ func (f Formatter) append(dst []byte, p *parser.Parser, params []interface{}) []
 		if name := string(p.ReadIdentifier()); name != "" {
 			if f.paramsMap != nil {
 				if param, ok := f.paramsMap[name]; ok {
-					dst = types.Append(dst, param, 1)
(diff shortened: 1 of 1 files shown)


vmihailenco:
This should be already possible without any changes:

```
shard.fmter.SetParam("shard", pg.Q(shard.Name())) // formatted as shard_9
shard.fmter.SetParam("shard", pg.F(shard.Name())) // formatted as "shard_9"
shard.fmter.SetParam("shard", shard.Name()) // formatted as 'shard_9'
```

PS Please make sure that you use latest go-pg and sharding packages.

liut:
Oh, I see.
Thank you.
Im also submitted a PR in sharding about `shard.SetParam()`

liut:
What a nice type `Q` :+1:

PR #179, opened 2016-04-13
Remove w.Close() from readCopyData()

dankeder:
The call to `w.Close()` prevents `db.CopyTo` from writing multiple dumps to stdout like this:

    db.CopyTo(os.Stdout, 'COPY mytable1 TO STDOUT')
    
    # the following will fail because os.Stdout is closed
    db.CopyTo(os.Stdout, 'COPY mytable2 TO STDOUT')

I'm not sure if there is some reason for the call to Close(), the commit that introduced it does not mention it, so please approach this PR with caution.

Also, I'm new to Golang and this is my first Golang PR, so please bear with me :-).

### DIFF
--- a/db.go
+++ b/db.go
@@ -205,7 +205,7 @@ func (db *DB) CopyFrom(r io.Reader, query interface{}, params ...interface{}) (*
 }
 
 // CopyTo copies data from a table to writer.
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks - makes sense!

PR #207, opened 2016-05-04
Release acquired DB connections in db.CopyTo()

dankeder:
Otherwise it will program runs out of pool connections and crashes with `pg: connection pool timeout`


vmihailenco:
Thanks, going to add tests.

vmihailenco:
Released v4.1.4

PR #208, opened 2016-05-04
Make it possible to disable connection pool rate limiting

dankeder:
When executing many queries in quick succession the connection pool will hit the threshold of the rate-limiter. This PR makes it possible to turn off the rate limiting in the connection pool.

The default behaviour is not changed - the rate limiting is turned on by default.

Example of use:

        db := pg.Connect(&pg.Options{
            Addr: "localhost:5432",
            User: "postgres"
            Password: "somepassword",
            Database: "template1",
            DisableRateLimiting: true,
        })


vmihailenco:
>When executing many queries in quick succession the connection pool will hit the threshold of the rate-limiter.

This can't/should not be true. When your queries don't return fatal errors (such as network timeouts) connections are reused and you can't hit rate-limit threshold. In other words if you are rate limited then there is something wrong with connection pool and/or Postgres setup.

Overall I see such PRs time to time and so far the root reasons were bugs in lib/user code.

dankeder:
It seems you are right, it's probably a bug somewhere. I created a small module to demonstrate the behaviour here:

https://gist.github.com/dankeder/33192df083703aedbaf5d4c42cb0f999

When calling `db.Query()` 1000 times it works properly. When calling `db.CopyTo()` 1000 times it crashes after 34 iterations (with my setup, ymmw).

Do you think it's something wrong with how `CopyTo()` is used? Or is it a bug in `CopyTo()` itself?

vmihailenco:
Can you please try latest master? It has at least 2 fixes for CopyFrom/CopyTo.

dankeder:
I tried the latest master, the problem seems to be gone. Thanks very much!

Issue #227, opened 2016-05-20
Use a db-prepared statement on a Tx


danilobuerger:
With the stdlib one can do:

```
tx.Stmt(someStmt)
```

to get a previously prepared statement for that specific Tx. Is something like that possible with go-pg?

vmihailenco:
I've just added similar API to go-pg https://godoc.org/gopkg.in/pg.v4#Tx.Stmt

PR #230, opened 2016-05-21
Add Tx.Stmt API.

vmihailenco:


vmihailenco:
Fixes https://github.com/go-pg/pg/issues/227

PR #376, opened 2016-10-25
Rename Order.

vmihailenco:
This is backwards incompatible but since release was only few days ago I think it is fine :)


vmihailenco:
And the plan is that I will add `Order("column1", "column2 DESC")` back and it will make sure that column name is properly escaped.

PR #385, opened 2016-10-31
Remove ratelimit.

vmihailenco:


kyrylo:
Curious to know why? I've been getting some `pg: you open connections too fast` messages and now I see this is gone. Interested only for education purposes.

vmihailenco:
I still think it is useful, but
- users don't understand how it works
- sometimes it is annoying when you really need to open and close a lot of connections (like in tests)
- none else does this
- there are more sophisticated ways to prevent app from opening too many connections

PR #431, opened 2016-12-12
Add SetParam(string, interface{}) for custom Formater.namedParams

liut:
I have many alike tables in database schema, and they need transaction when operated them, like:
```sql
CREATE TABLE  alias_email (LIKE alias_ INCLUDING ALL);
CREATE TABLE  alias_mobile (LIKE alias_ INCLUDING ALL);
CREATE TABLE  alias_credit (LIKE alias_ INCLUDING ALL);
CREATE TABLE  alias_idcard (LIKE alias_ INCLUDING ALL);

CREATE TABLE device_android(LIKE device INCLUDING ALL, CHECK(cate_id = 1));
CREATE TABLE device_ios(LIKE device INCLUDING ALL, CHECK(cate_id = 2));

```
code sample
```go
type Alias struct {
	TableName string `sql:"alias_?SHARDA" json:"-"`
}

type Device struct {
	TableName string `sql:"device_?SHARDD" json:"-"`
}
```

### DIFF
--- a/db.go
+++ b/db.go
@@ -66,6 +66,11 @@ func (db *DB) WithParam(param string, value interface{}) *DB {
 	return &cp
 }
 
+// Allow custom namedParams
+func (db *DB) SetParam(key string, value interface{}) {
+	db.fmter.SetParam(key, value)
+}
+
 func (db *DB) conn() (*pool.Conn, error) {
 	cn, _, err := db.pool.Get()
 	if err != nil {
--- a/tx.go
+++ b/tx.go
@@ -302,3 +302,8 @@ func (tx *Tx) CopyFrom(r io.Reader, query string, params ...interface{}) (*types
 	tx.freeConn(cn, err)
 	return res, err
 }
+
+// Allow custom namedParams
+func (tx *Tx) SetParam(key string, value interface{}) {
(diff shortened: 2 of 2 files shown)


vmihailenco:
There is already `WithParam` API - https://github.com/liut/pg/blob/4eb4a09a3dd6ef612b6dea8389fd7927eb110401/db.go#L58-L67. It is used by pgsharding - https://github.com/go-pg/sharding/blob/v5/cluster.go#L62-L67

liut:
@vmihailenco `WithParam` can not use in transaction with multi-table in my case:
```go
nextId := 'select nextval(some_seq_id)'
tx.SetParam('SHARDA', pg.Q(some_key_suffix_from_nextId))
orm.Insert(tx, objA)
objB.A_Id = objA.Id
tx.SetParam('SHARDB', pg.Q(some_key_with_obj_A))
orm.Insert(tx, objB)
```

vmihailenco:
Is that a real use case? Suggested API is rather ugly and cross-shard transactions look wrong...

liut:
@vmihailenco Yes, you right, It is not pretty good. I'm looking for better solution again.

PR #435, opened 2016-12-13
Add `ieq` : `ILIKE` and `match` : `SIMILAR TO` in `url_values`

liut:
- other fixes for order empty string


vmihailenco:
Looks good. Thanks!

PR #438, opened 2016-12-16
Custom error logger and query logger

liut:
Use interface instead of `*log.Logger`, remove dependency of go `log`

### DIFF
--- a/internal/log.go
+++ b/internal/log.go
@@ -2,16 +2,23 @@ package internal
 
 import (
 	"fmt"
-	"log"
 	"path/filepath"
 	"runtime"
 	"strings"
 )
 
 var (
-	Logger      *log.Logger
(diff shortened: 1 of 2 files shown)


internal/log.go
```diff
@@ -2,17 +2,24 @@ package internal
...
+	QueryLogger IQueryLogger
 )
 
+type IErrorLogger interface {
+	Output(calldepth int, s string) error
+}
+
+type IQueryLogger interface {
+	Printf(format string, v ...interface{})
+}
```
vmihailenco: Can you please merge IErrorLogger and IQueryLogger into 1 `internal.Logger` interface? Thanks

liut: Sure, that is a better idea.

PR #458, opened 2017-01-16
InsecureSkipVerify and default database user name

szubtsovskiy:
To make `ParseURL` working in Heroku I've changed default TLSConfig created when parsing database URL to have `InsecureSkipVerify=true` . This is a minimum for this to work in Heroku.

I've also set default database user name to 'postgres' because otherwise Postgres is throwing an error (not a case for Heroku but good to have).


vmihailenco:
Thanks, LGTM

PR #460, opened 2017-01-17
Update README.md

thinkclay:
Breaking up massive README into Wiki pages for easier collab and management


thinkclay:
In addition to breaking documentation out into a wiki, do you think it makes sense to do an `examples` folder in the repo to show things like migrations and mini working apps, or would you rather those all live in the wiki as well?

vmihailenco:
Thanks - I already like it. Some pages (like installation) are rather small and probably don't deserve separate wiki page, but we can merge it later I guess. Don't hesitate to make such changes if it makes sense :)

>to show things like migrations and mini working apps, or would you rather those all live in the wiki as well?

I don't have a preference so do whatever makes more sense. It should be very helpful anyway.

PR #481, opened 2017-02-27
Add WhereIn orm method

rodzzlessa24:


vmihailenco:
Looks good. Could you add a test or example?

rodzzlessa24:
Sorry the method does not work I was mistakenly testing it with my old WhereIn method.

Issue #509, opened 2017-04-01
IF NOT EXISTS Option for Create Table


NickPancakes:
I would like to suggest a new option for orm.CreateTableOptions that will use "IF NOT EXISTS" syntax when creating tables. This is useful in my application as I attempt to initialize all tables using CreateTable on application start. I can only assume it would be equally as useful for others.

NickPancakes:
Added PR of my own implementation. I hope it's an acceptable request and solution!

fangker:
think it's  useful advice

PR #510, opened 2017-04-01
orm: Add IgnoreExists option for CreateTable

NickPancakes:
Closes go-pg/pg#509

### DIFF
--- a/orm/create_table.go
+++ b/orm/create_table.go
@@ -3,8 +3,8 @@ package orm
 import "errors"
 
 type CreateTableOptions struct {
-	Temp         bool
-	IgnoreExists bool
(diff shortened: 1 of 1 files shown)


orm/create_table.go
```diff
@@ -3,7 +3,8 @@ package orm
 import "errors"
 
 type CreateTableOptions struct {
-	Temp bool
+	Temp         bool
+	IgnoreExists bool
```
vmihailenco: Please rename to `IfNotExists` so it matches PostgreSQL naming

NickPancakes: Done.

vmihailenco:
:+1:

PR #516, opened 2017-04-13
Fix/order by nulls first

rcmonitor:
When using PostgreSQL 'NULLS FIRST' order modifier, it should take only the column name in quotes:

`"column_name" DESC NULLS FIRST`
, not all the ORDER BY expression as it was before:

`"column_name DESC NULLS FIRST"`

test fixed accordingly


vmihailenco:
Thanks. Just a hint that you can/should use `q.OrderExpr()` when you need complex order in cases like this.

PR #579, opened 2017-07-17
Fix create table query generation

zyukin:


orm/create_table.go
```diff
@@ -46,8 +46,9 @@ func (q createTableQuery) AppendQuery(b []byte) ([]byte, error) {
 	b = append(b, " ("...)
 
 	for i, field := range table.Fields {
+		b = append(b, `"`...)
 		b = append(b, field.SQLName...)
```
vmihailenco: Can you try `field.Column`? I think it should be already escaped.

zyukin: It's better!

PR #599, opened 2017-07-28
fix connection leak in case tx begin error

quexer:
connection should be closed if "tx.Exec("BEGIN")" fail


vmihailenco:
Thanks!

PR #616, opened 2017-08-14
Fix order by error

kazhuravlev:
reproduce err:
```
db.Model(&MyModel{}).Order("field1").First()

// ==>  select * from my_model order by field1, limit 1;
```

### DIFF
--- a/orm/select.go
+++ b/orm/select.go
@@ -105,7 +105,7 @@ func (q selectQuery) AppendQuery(b []byte) ([]byte, error) {
 		if len(q.q.order) > 0 {
 			b = append(b, " ORDER BY "...)
 			for i, f := range q.q.order {
-				if i > 0 {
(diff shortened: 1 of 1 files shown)


kazhuravlev:
how about strings.Join() or bytes.Join() ?

kazhuravlev:
warn! not tested other scenarios

vmihailenco:
The problem here is that `MyModel` does not have primary key that is used in `First` to sort table. I think the correct fix is to return an error when pk is not found.

kazhuravlev:
I have tables without pk. There may be no pk. It is normal.

vmihailenco:
Then you should use `Limit(1)` instead of `First`, because `First` is supposed to sort table.

kazhuravlev:
oh, I misunderstood you. Thanks!

PR #682, opened 2017-11-17
add context.Context support for transactions

### LINKED ISSUE #681: Add Context() method to type Tx
Unlike type DB, there doesn't appear to be an easy way to get the attached context from a transaction right now. We would like to have a way to do so for logging purposes in our OnQueryProcessed handler.

bgardiner:
This PR resolves #681.

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -1,6 +1,7 @@
 package pg
 
 import (
+	"context"
 	"errors"
 	"io"
 	"time"
@@ -298,4 +299,14 @@ func (tx *Tx) close(lastErr error) error {
 	tx.cn = nil
 
 	return err
+}
+
+func (tx *Tx) Context() context.Context {
+	return tx.db.Context()
+}
+
+func (tx *Tx) WithContext(ctx context.Context) *Tx {
+	cp := *tx
+	cp.db = cp.db.WithContext(ctx)
+	return &cp
 }


vmihailenco:
Thanks

vmihailenco:
Hi,

I am sorry, but I had to remove `tx.WithContext` in https://github.com/go-pg/pg/pull/692. The main reason is that now `Tx` has `sync.Mutex` which can not be copied. It is also kinda strange that following code shares the same connection, but can be closed separately:

```
tx1, _ := db.Begin()
tx2 := tx1.WithContext(ctx)
// now tx1 and tx2 share the same connection and state which is awkward
```

Hopefully you can replace code above with:

```
tx1, _ := db.Begin()
tx2, _ := db.WithContext(ctx).Begin()
```

bgardiner:
Thanks. Yes, `tx.Context` was the primary functionality we needed so we should be good.

maxatome:
How about adding `Context()` to the `orm.DB` interface?
Just to avoid checking types behind `orm.DB` just to retrieve the `context.Context`, from a `pg.QueryProcessedEvent` for exemple...

maxatome:
`DB.WithParam()` and `DB.WithTimeout()` discard `ctx`. Is it a desired behavior?

It is a bit confusing that `db.WithParam(…).WithContext(ctx)` does not equal `db.WithContext(ctx).WithParam(…)`. In the second case, the context is lost…

PR #687, opened 2017-11-30
types: implement values operator

siteshen:


vmihailenco:
Can you add an example how this is going to be used?

siteshen:
@vmihailenco I need to fetch rows by composite keys like this, and it's tedious and error-prone to concat sql:
```sql
SELECT * FROM "storages" WHERE ((type, name) IN (VALUES ('sql','postgres'), ('nosql','redis')));
```
And i wrote a runnable demo:
```go
package main

import (
	"github.com/go-pg/pg"
	"github.com/go-pg/pg/orm"
	"github.com/go-pg/pg/types"
)

type Storage struct {
	Type string
	Name string
}

func main() {
	// shell: createdb demo
	db := pg.Connect(&pg.Options{
		Database: "demo",
		User:     "",
	})

	if err := db.CreateTable(&Storage{}, &orm.CreateTableOptions{IfNotExists: true}); err != nil {
		panic(err)
	}

	storages := []*Storage{}
	if err := db.Model(&storages).WhereIn("(type, name) IN (?)", types.Values([]interface{}{
		[]string{"sql", "postgres"},
		[]string{"nosql", "redis"},
	})).Select(); err != nil {
		panic(err)
	}
	// sample output:
	// SELECT * FROM "storages" AS "storage" WHERE ((type, name) IN (VALUES ('sql','postgres'), ('nosql','redis')));
}

```

vmihailenco:
Please update go-pg and try following:

```
	err := db.Model(&storages).
		WhereIn("(type, name) IN (?)",
			[]string{"sql", "postgres"},
			[]string{"nosql", "redis"}).
		Select()
```

It should do the same with more concise syntax/API.

siteshen:
Thanks, it works.

PR #690, opened 2017-11-30
Support multiple values in pg.In

vmihailenco:
Replaces https://github.com/go-pg/pg/pull/687

### DIFF
--- a/types/in_op.go
+++ b/types/in_op.go
@@ -6,33 +6,53 @@ import (
 )
 
 type InOp struct {
-	slice  reflect.Value
-	append AppenderFunc
+	slice reflect.Value
 }
 
 var _ ValueAppender = (*InOp)(nil)
 
-func In(slice interface{}) *InOp {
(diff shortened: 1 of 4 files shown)


types/in_op.go
```diff
@@ -6,33 +6,53 @@ import (
...
 
-func In(slice interface{}) *InOp {
-	v := reflect.ValueOf(slice)
+func In(values ...interface{}) *InOp {
+	var v reflect.Value
+	if len(values) == 1 {
+		v = reflect.ValueOf(values[0])
+	} else {
+		v = reflect.ValueOf(values)
+	}
```
siteshen: This breaks `Query.WhereIn()` API:
```go
query := orm.Query{}
query.WhereIn("id in(?)", 1) // panic: pg.In(unsupported int)
```

vmihailenco: Should be fixed by https://github.com/go-pg/pg/pull/693

PR #693, opened 2017-12-01
Fix pg.In for slices with 1 element

vmihailenco:


siteshen:
Excute me, but it fails sometimes
```go
Query.WhereIn("(type, name) IN (?)", []string{"sql", "postgres"})

// will generate sql
`select * from table where (type, name) in ("sql", "postgres") -- wrong`
// instead of
`select * from table where (type, name) in (("sql", "postgres")) -- correct`
```

Of course, i can "correct" it like this, but it seems a bug of `pg.WhereIn` API
 ```go
Query.WhereIn("(type, name) IN (?)",[][]string{[]string{"sql", "postgres"}})
```

The problem is `types.In` can't distinct the queries:
```go
Query.WhereIn("(type, name) IN (?)", []string{"sql", "postgres"})
// where (type, name) in (("sql", "postgres"))

Query.WhereIn("name IN (?)", []string{"postgres", "mysql"})
// where name in ('postgres', 'mysql')
```

I think a better way to solve it is to restore `types.In` api (like The Zen of Python: "There should be one-- and preferably only one --obvious way to do it."), and make `inOp.AppendValue()` support nested type (seems function `appendIn` already support nested slice).
```go
func In(slice interface{}) ValueAppender {}
```
The element of lice may be simple type like integer, string, or complex type like slice of int, slice of string, even more complex type like slice of slice of slice of string.

And the client sample code:
```go
package main

import (
	"github.com/go-pg/pg"
	"github.com/go-pg/pg/orm"
)

type Storage struct {
	ID   int64
	Type string
	Name string
}

func main() {
	db := pg.Connect(&pg.Options{
		Database: "demo",
		User:     "",
	})
	if err := db.CreateTable(&Storage{}, &orm.CreateTableOptions{IfNotExists: true}); err != nil {
		panic(err)
	}

	storages := []*Storage{}
	db.Model(&storages).WhereIn("(type, name) IN (?)", []string{"sql", "postgres"}).Select()
	db.Model(&storages).WhereIn("name IN (?)", "postgres", "mysql").Select()
}
```

vmihailenco:
The problem here is that I wanted to support old `pg.In` behaviour for cases like `Where("id in (?)", pg.In([]int{1, 2, 3}))`. But apparently it causes more problems than it solves...

siteshen:
`types.In` have nothing to do with `Where("id in (?)", ValueAppender{})`, if `type.inOp` impl ValueAppender correctly,  `Where()` shoud works.

So what we need to do is defining `types.inOp`'s behavior clearly.

The origin `types.InOp` (before #690) support simple type (string/int) only
```javascript
In(["a1", "a2"]) // => ('a1', 'a2')
```

We just make it support nested types, and everything should work.

```javascript
In([["a1", "b1"], ["a2", "b2"]]) // => (('a1', 'b1'), ('a2', 'b2'))
In([["a1", 1], ["a2", 2]]) // => (('a1', 1), ('a2', 2))
```
BTW, if slice element implement interface `ValueAppender`, `types.inOp` should follow it.

siteshen:
Update  last comment, remove outer parentheses:

```javascript
// element is simple type
In(["a1", "a2"]) // 'a1', 'a2'
In([1, 2]) // 1, 2
```

```javascript
// element is slice of interface{}
In([["a1", "b1"], ["a2", "b2"]]) // ('a1', 'b1'), ('a2', 'b2')
In([["a1", 1], ["a2", 2]]) // ('a1', 1), ('a2', 2)
```

vmihailenco:
Please try latest master. `pg.In` is reverted to the old behavior. `pg.InMulti` is added to support multiple values. `Query.WhereIn` should continue to work.

siteshen:
thanks, it works.

PR #734, opened 2018-01-16
Use UTC timezone when Postgres type does not have timezone info

vmihailenco:


johnnybravo-xyz:
hi @vmihailenco 
can you tell me why this change was done?
I was expecting go-pg to not change `Location` for `time.Time` instances when dealing with `date` columns of postgres which does not have timezone information, I found this PR relevant:
https://github.com/go-pg/pg/pull/6

vmihailenco:
AFAIR using `time.UTC` works better (saves time correctly) with `timestamp` and `date` types since PostgreSQL ignores timezone for such times.

PR #744, opened 2018-01-24
jmk - Enuser connection is freed on panic

7sedam7:


vmihailenco:
I know that database/sql does something similar, but I am skeptical because:
- it changes the way you write code - you assume that panic is perfectly normal and defer is the way to go
- still leaves `*pg.DB` in unknown state, because you don't know at which point panic occurred and connection is put to the pool no matter what

On the contrary panic causes app to exit and be restarted by systemd/upstart. This seems to be a safer strategy for a web app in a case of such exceptional situation such as panic.

Did you really encounter a panic inside `simpleQueryData` function? If so please file an issue and I will try to fix it.

PS What is jmk? :)

7sedam7:
I'm encountering a problem that connections to pg are getting opened more and more (number of connections is growing), I still have no idea why, but as far as I see, I don't even have control over opening and closing connections with using `go-pg`. 

I have one singleton instance of `*pg.DB` and use it in the whole app.

This was an attempt to fix it. If you have some ideas how/why could my problem happen please share your thought :)

`jmk` are my initials :p
@vmihailenco

vmihailenco:
>If you have some ideas how/why could my problem happen please share your thought :)

Do you use `Tx` or `Stmt`? If so please that `Tx` is committed or rollbacked. `Stmt` should be closed after it is not needed...

7sedam7:
I use `Tx`, but it seems I had to do a manual rollback on error, while I expected that rollback happens automatically (and in the database it was by database mechanism of failed  transaction), but new connection was opened, and old one stays hanging.

vmihailenco:
Correct, I had an idea to improve it in https://github.com/go-pg/pg/issues/725, but it would mean that go-pg does not even send queries when `Tx` is auto-rollbacked on error, .e.g.

```
tx := db.Begin()
tx.Exec("QUERY 1") // error happens, tx is rollbacked, conn is closed
tx.Exec("QUERY 2") // this query is not sent to the server at all, because there is no tx and/or conn
```

I decided that it would be too magical and can potentially break some code/tricks. Instead it seems much easier to always use defer:

```
tx := db.Begin()
defer tx.Rollback()

tx.Exec("QUERY 1") 
tx.Exec("QUERY 2")
tx.Commit()
```

7sedam7:
Ok, one suggestion to mention it very explicitly, people might expect different behavior :)

Issue #754, opened 2018-02-02
pg: can't find column=... in model=... when using pg.Query


trietphm:
I have this model

```go
type User struct {
  id int
  name string
}
```
And in the database, the table `users` is:

```
        Column         |           Type           |
-----------------------+--------------------------+
 id                    | integer                  |
 name                  | character varying(100)   |
 age                   | integer                  |
... <about 40 columns>
-----------------------+--------------------------+
```

I know that I can query the data using `Model` like this

```go
var user User
err := db.Model(&user).Where("id = ?", id).Select()
```
`pg` will generate query `SELECT users.id, users.name FROM users WHERE id = ?` and ignore column `age` because `age` is not defined in struct `User`.

But for some reason, I want to write raw query `SELECT * FROM users WHERE id = ?`. 

```go
var user User
sql:= "SELECT * FROM users WHERE id = ?"
_, err := db.Query(&user, sql, id)
```
And above code will return error: `pg: can't find column=age in model=User`

I don't want to defined all fields in struct `User` or have to write all needed columns in query `SELECT`, is there anyway to handle that? Or can I write a custom scanner for my model?

Thank you.

vmihailenco:
One possible solution would be to add a model tag/option like `,discard_unknown_columns` that will make go-pg to return `nil` error instead of `pg: can't find column=age in model=User`. Would that solve your problem?

vmihailenco:
Should be done - see https://godoc.org/github.com/go-pg/pg#example-DB--DiscardUnknownColumns

trietphm:
Cool! Thank you.

gadelkareem:
In case of a typo with column name in a query, this could get very buggy! is there an option to allow unknown columns per query? maybe add them to func `Query.IgnoreColumn("inner","outer"...)`

sanjaypsmc:
@vmihailenco This should either be the default behavior of go-pg or there should be big red warning label in capital letters in the wiki and README that go-pg isn't compatible with zero downtime migrations. It's entirely unexpected that an ORM would be so fundamentally incompatible with modern DB migration paradigms. This design decision alone means go-pg has the potential to be extremely destructive in production environments.

PR #755, opened 2018-02-02
orm: add discard_unknown_columns option

vmihailenco:
Fixes https://github.com/go-pg/pg/issues/754


sanjaypsmc:
@vmihailenco This does not support `Returning("*")` on Inserts. Can anything done about that?

vmihailenco:
Please create an issue with example that does not work.

Issue #496, opened 2017-03-20
Custom table name not quoted in queries


jack-chung:
Hello everyone,

I have a model like this:
```go
type User struct {
	tableName          struct{} `sql:"user"`
	ID                 int64
}
```
and db.Insert(&user) is generating queries without quoting the table name, resulting in an error due to reserved keyword. Surprisingly, if I don't define a table name override, by default it quotes the table name. It only skips the quotes when it's an override.

Is there any way to force the quotes?

jack-chung:
Ah, I figured out a workaround, by adding quotes into the tag like this: `sql:"\"user\""`.
The current behavior should probably still be considered a bug?

vmihailenco:
I think this is a bug, but I need to test that it does not break placeholders in case like `sql:"?shard.user"`. In case it works I will fix this in master (not v5).

vmihailenco:
But it definitely will break cases like:

```
type AdminUser struct {
    tableName struct{} `sql:"(select * from users where admin) as users"`
}
```

Not that I am using something like that, but...

PR #777, opened 2018-02-22
orm: consolidate isPostgresKeyword

0rax:
I encountered an issue with a table called `group` which triggered a syntax error from postgres (related to https://github.com/go-pg/pg/issues/496) as `GROUP` is a reserved keyword.

To fix that I checked the list of reserved keyword from [Postgres Documentation](https://www.postgresql.org/docs/current/static/sql-keywords-appendix.html) and added what would be possible table name to the check.

I also added a `strings.ToLower` call to the checked word as the check did not cover the case where a table is called `User` which can happen with database created by some previous ORM or a user.

### DIFF
--- a/orm/table.go
+++ b/orm/table.go
@@ -498,10 +498,11 @@ func (t *Table) newField(f reflect.StructField, index []int) *Field {
 
 func isPostgresKeyword(s string) bool {
 	switch s {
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks

PR #790, opened 2018-03-12
Omit zero value on notnull fields with default value

CrushedPixel:
Currently, it is not possible to create `NOT NULL` columns with `DEFAULT` values using `go-pg`,
because the `notnull` option causes zero values which otherwise represent `NULL` to be inserted in the database:

```go
type A struct {
	Id   int64
	Time time.Time `sql:",notnull,default:NOW()"`
}

func main() {
	// [... database initialization ...]
	a := &A{}
	if err := db.Insert(a); err != nil {
		panic(err)
	}
	log.Printf("%d %s\n", a.Id, a.Time)
	// outputs: 1 0001-01-01 00:00:00 +0000 UTC
}
```

To let `go-pg` determine whether the zero value should be inserted or `DEFAULT` should be used, we can use **pointer types**:
```go
type B struct {
	Id   int64
	Time *time.Time `sql:",notnull,default:NOW()"`
}

func main() {
	// [... database initialization ...]
	b := &B{}
	if err := db.Insert(b); err != nil {
		panic(err)
	}
	log.Printf("%d %s\n", b.Id, b.Time)
	// outputs: 1 2018-03-12 10:32:26.092161 +0000 +0000
}
```

If `notnull` is set for a field, but the value is a `nil` pointer, `DEFAULT` is used when inserting.

### DIFF
--- a/orm/field.go
+++ b/orm/field.go
@@ -71,16 +71,8 @@ func (f *Field) IsZero(strct reflect.Value) bool {
 }
 
 func (f *Field) OmitZero(strct reflect.Value) bool {
-	return !f.HasFlag(NotNullFlag) && f.isZero(f.Value(strct))
-}
-
-// InsertDefault returns whether this field should be set to DEFAULT
-// when inserting strct.
-// This is the case if either the field is nullable and has its zero value
-// or the field is not nullable and the value is a nil pointer.
-func (f *Field) InsertDefault(strct reflect.Value) bool {
 	val := f.Value(strct)
-	return f.OmitZero(strct) ||
+	return (!f.HasFlag(NotNullFlag) && f.isZero(val)) ||
 		(f.HasFlag(NotNullFlag) && val.Kind() == reflect.Ptr && val.IsNil())
 }
 
--- a/orm/field.go
+++ b/orm/field.go
@@ -74,6 +74,16 @@ func (f *Field) OmitZero(strct reflect.Value) bool {
 	return !f.HasFlag(NotNullFlag) && f.isZero(f.Value(strct))
 }
 
+// InsertDefault returns whether this field should be set to DEFAULT
+// when inserting strct.
+// This is the case if either the field is nullable and has its zero value
+// or the field is not nullable and the value is a nil pointer.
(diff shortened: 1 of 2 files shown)


CrushedPixel:
I realized that we should omit `nil` pointers for `notnull` fields when updating values in the database to avoid accidental `NOT NULL` constraint violations.

vmihailenco:
Why not something like

```
func (f *Field) OmitZero(strct reflect.Value) bool {
	return f.Default != "" ||
		(!f.HasFlag(NotNullFlag) && f.isZero(f.Value(strct)))
}
```

Such fix perfectly matches the description of the problem.

CrushedPixel:
In your code, `OmitZero` would always return `true` if a default value is set, causing the actual field values to always be ignored by the following code ([`orm/update.go#L222-L226`](https://github.com/go-pg/pg/blob/master/orm/update.go#L222-L226)):
```go
if f.OmitZero(v) {
	b = append(b, "NULL"...)
} else {
	b = f.AppendValue(b, v, 1)
}
```

This means that you couldn't set the values of fields with default value anymore.

vmihailenco:
Okay, so it should be `(f.Default != "" || !f.HasFlag(NotNullFlag)) && f.isZero(f.Value(strct))`

gadelkareem:
I think this broke the `,notnull` functionality somehow. If the field has a comment `,notnull` alone then it works normally. But if I add `,notnull,default:-1` for ex, then the query is sent with null values!
Getting error:
```
ERROR #23502 null value in column "total" violates not-null constraint 
```

vmihailenco:
Where is the code?

gadelkareem:
something like this
```
type Book struct{
    total        int64             `sql:"total,notnull,default:-1,type:BIGINT"` 
}
PgDb.Model(&updatedBooks).Update()

```

abronan:
I just encountered that error on a bulk update. I had defined both `,notnull,default:false` and got `null value in column "XXX" violates not-null constraint `.

I had to remove the `default:false` (but this is was enforced in the schema anyway). This seems to work properly now.

PR #819, opened 2018-04-07
Add HasTable

thanhps42:


orm/has_table.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	if q.q.model == nil {
+		return nil, errors.New("pg: Model(nil)")
+	}
+
+
+	b = append(b, fmt.Sprintf(
+		"SELECT EXISTS (SELECT 1 FROM pg_tables WHERE schemaname = 'public' AND tablename = %s);", strings.Replace(
+			string(q.q.model.Table().Name), `"`, "'", -1),
+	)...)
```
vmihailenco: Please get rid of `fmt.Sprintf` and use append as in other places:
- for consistency
- performancy
- readability (code becomes more readable when your query is bigger and more complex)

vmihailenco:
Also please add simple test or example using sequence of `HasTable`, `CreateTable` and `HasTable`.

thanhps42:
Thanks for your comment. I will fix it now (y)

PR #821, opened 2018-04-08
Add HasTable and HasColumn

thanhps42:


orm/query.go
```diff
@@ -812,6 +812,46 @@ func (q *Query) DropTable(opt *DropTableOptions) (Result, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	if count > 0 {
+		return true, nil
+	}
+
+	return false, nil
+}
```
vmihailenco: Let's not add these 2 methods to the `*Query` since it clutters API and is not very useful, e.g. `db.Model(&strct).Where().HasTable` does not make much sense. You should be able to move implementation from `Query.HasTable` to `orm.HasTable` without any issues...

orm/has_table_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+var _ = Describe("HasTable", func() {
+	It("has table", func() {
+		q := NewQuery(nil, &HasTableModel{})
+
+		b, err := hasTableQuery{q: q}.AppendQuery(nil)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(b)).To(Equal("SELECT count(*) FROM pg_tables WHERE schemaname = 'public' AND tablename = 'has_table_models';"))
+	})
+	It("has table with custom name", func() {
```
vmihailenco: separatetests with newline

orm/has_table.go
```diff
@@ -0,0 +1,41 @@
...
+		return nil, errors.New("pg: Model(nil)")
+	}
+
+	tableName := string(q.q.appendTableName(nil))
+	tableName = strings.TrimPrefix(tableName, `"`)
+	tableName = strings.TrimSuffix(tableName, `"`)
+
+	b = append(b, "SELECT count(*) FROM pg_tables WHERE schemaname = 'public' AND tablename = '"...)
+	b = append(b, tableName...)
+	b = append(b, "';"...)
```
vmihailenco: No need for `;` in the end

orm/query.go
thanhps42: I think CreateTable and DropTable should be move too.

vmihailenco:
Sorry, but I've just reverted this PR because this API is too specific and inflexible (and was never released). Better API would be `db.DescribeTable(model interface{}) ...` that returns some data structure with list columns with their types and so on. Not only it is one method instead of 2 but it provides more data and is more broadly useful.

PR #849, opened 2018-05-16
update options URI scheme

kxjhlele:
The URI scheme designator can be either postgresql:// or postgres://. Each of the URI parts is optional. The following examples illustrate valid URI syntax uses:


vmihailenco:
Okay, please fix tests.

vmihailenco:
Replaced by https://github.com/go-pg/pg/pull/853, but thanks for bringing this anyway.

PR #853, opened 2018-05-16
Accept postgresql as scheme.

timpwbaker:
This commit updates the options.go file to accept postgresql as a
scheme.

As per the postgresql docs the scheme can either be postgres:// or
postgresql://

    The general form for a connection URI is:
    postgresql://[user[:password]@][netloc][:port][/dbname][?param1=value1&...]
    The URI scheme designator can be either postgresql:// or postgres://. Each of the URI parts is optional.

It also updates a the test to include this additional option

### DIFF
--- a/options.go
+++ b/options.go
@@ -134,7 +134,7 @@ func ParseURL(sURL string) (*Options, error) {
 	}
 
 	// scheme
-	if parsedUrl.Scheme != "postgres" {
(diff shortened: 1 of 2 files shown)


timpwbaker:
Thank you for all the great work.

I realised after submitting this there's another PR for the same thing that's failing CI. We're currently blocked by this so I thought I'd submit a green PR.

PR #907, opened 2018-07-16
Format order with params

### LINKED ISSUE #902: Cannot use TableAlias in .Order()
`.Order("?TableAlias.id")` doesn't substitute alias in place of ?TableAlias.
`.OrderExpr("?TableAlias.id")` works just fine but I believe they both should format query with param substitutions.

23doors:
fixes #902

### DIFF
--- a/orm/query.go
+++ b/orm/query.go
@@ -470,7 +470,7 @@ loop:
 			switch internal.UpperString(sort) {
 			case "ASC", "DESC", "ASC NULLS FIRST", "DESC NULLS FIRST",
 				"ASC NULLS LAST", "DESC NULLS LAST":
-				q = q.OrderExpr("? ?", types.F(field), types.Q(sort))
+				q = q.OrderExpr(field+" ?", types.Q(sort))
 				continue loop
 			}
 		}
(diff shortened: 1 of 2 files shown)


23doors:
@vmihailenco  I just started using go-pg recently so I'm not sure if this approach is completely ok. Correct me if I'm wrong.

vmihailenco:
Why not to use `OrderExpr`? `Order` is supposed to be as strict as possible so you can feed it any string and it produces safe SQL.

23doors:
Sure, OrderExpr works but it just seems that param evaluation should be standard behavior. At least for such basic param like ?TableAlias which is the only way to avoid having ambigous columns in multi-table queries. 

It’s just strange to me and it’s nor exactly anywhere in docs that params works in Where() but not in Order().

23doors:
But I think I see what you're getting at. Maybe consider merging .Order() and .OrderExpr()? Or add a note to .Order() and .Columns() that params are not evaluated there?

vmihailenco:
Yeah, I guess we need a note. Do you care to send a PR? :)

23doors:
#912

PR #908, opened 2018-07-16
Fix formatting structs in Where with params

### LINKED ISSUE #903: When TableAlias is used in parenthesis, param is not evaluated
`q.Where("?TableAlias = ?", "param")` => works fine. TableAlias is substituted and param converted to sql Value.

`q.Where("(?TableAlias) = ?", "param")` => TableAlias is substituted but `?` is not evaluated. Results in e.g. `(table) = ?`

This is troublesome when using pg functions e.g.:
`q.Where("some_pg_func(?TableAlias.field) > ?", param)`

23doors:
fixes #903

### DIFF
--- a/orm/format.go
+++ b/orm/format.go
@@ -230,10 +230,7 @@ func (f Formatter) append(dst []byte, p *parser.Parser, params []interface{}) []
 			if !namedParamsOnce && len(params) > 0 {
 				namedParamsOnce = true
 				if len(params) > 0 {
-					tableParams, ok = newTableParams(params[len(params)-1])
-					if ok {
-						params = params[:len(params)-1]
-					}
+					tableParams, _ = newTableParams(params[len(params)-1])
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


orm/select_test.go
```diff
@@ -178,6 +181,16 @@ var _ = Describe("Select", func() {
...
 		Expect(string(b)).To(Equal(`SELECT * WHERE (TRUE)`))
 	})
+
+	It("expands ?TableAlias in Where with structs", func() {
+		t := time.Date(2006, 2, 3, 10, 30, 35, 987654321, time.UTC)
+		q := NewQuery(nil, &SelectModel{}).Column("id").Where("?TableAlias.name > ?", t)
+
+		b, err := selectQuery{q: q}.AppendQuery(nil)
+		Expect(err).NotTo(HaveOccurred())
+		fmt.Println(string(b))
```
vmihailenco: remove fmt.Println

orm/format.go
```diff
@@ -230,10 +230,7 @@ func (f Formatter) append(dst []byte, p *parser.Parser, params []interface{}) []
 			if !namedParamsOnce && len(params) > 0 {
 				namedParamsOnce = true
 				if len(params) > 0 {
```
vmihailenco: Looks like this check is not needed, because len is already checked on line 230. Can you remove it?

23doors:
@vmihailenco fixed!

vmihailenco:
Thanks!

PR #912, opened 2018-07-17
Add info about param expansion

### LINKED ISSUE #902: Cannot use TableAlias in .Order()

23doors:
fixes #902


vmihailenco:
Thanks!

PR #920, opened 2018-07-22
Add sslmode=required support in ParseURL()

### LINKED ISSUE #919: sslmode=required not supported / ParseURL
Using a connection string with  `...&sslmode=required` yields a "not supported" error from ParseURL().

louis77:
Resolves #919


vmihailenco:
Where these options come from? In https://www.postgresql.org/docs/10/static/libpq-ssl.html it is called `require`, not `required`.

Could you also add a test to the https://github.com/go-pg/pg/blob/master/options_test.go#L10?

PR #948, opened 2018-08-12
Add Options.MinIdleConns

vmihailenco:


jceim:
This option does not work for me when passed into `pg.Connect`. I think it's because it is not propagated to `pool.NewConnPool`. Same for the option `MaxConnAge`...

vmihailenco:
Thanks, should be fixed by https://github.com/go-pg/pg/pull/1130

PR #970, opened 2018-09-05
fix soft delete when deleted_at column is mapped to *time.Time

lanre-ade:
Soft delete fails when `deleted_at` column is mapped to a struct field of type `*time.Time`
`Panic: reflect.Set: value of type time.Time is not assignable to type *time.Time`


vmihailenco:
Thanks!

PR #1012, opened 2018-10-26
Add go mod files

jqs7:


vmihailenco:
Sorry, but we don't support go mod yet

jqs7:
¯\_(ツ)_/¯

mewwts:
I would love to be able to pin the version go-pg using go mod.

Issue #1043, opened 2018-11-19
OpenCensus support


edevil:
I'm a go-pg and OpenCensus user, and am looking for the best way to introduce tracing in go-pg queries. AFAIK, the database/sql support cannot be used with go-pg since it does not export that interface. I was looking at the hooks available, but I could only find per-model hooks.

Are there some global hooks that I can use for every query, in which I would be able to access a supplied context?

edevil:
@vmihailenco Would you be willing to consider a `OnQueryStart` hook, dual to `OnQueryProcessed`, so that we could start the tracing span there, store it in the context, and then retrieve the span from the context and End() it at OnQueryProcessed?

edevil:
It’s done.

vmihailenco:
:rocket:

giautm:
Hi, I want to intergate Opencensus with go-pg. But dont know how to deliver, can you provide an example?

edevil:
@giautm You need to add a query hook that uses `trace.StartSpan()` in `BeforeQuery()` and then closes the span on `AfterQuery()`.

giautm:
@edevil I don’t know how to get query (main table, fields, params) from the hook. It’s just a formatted string.

edevil:
@giautm Yes, I think you would need to parse the query in order to separate those attributes besides the params.

PR #1047, opened 2018-11-23
Add OnQueryStarted hook

edevil:
Addresses #1043

### DIFF
--- a/hook.go
+++ b/hook.go
@@ -96,6 +96,7 @@ func (db *DB) queryStarted(
 		Query:   query,
 		Params:  params,
 		Attempt: attempt,
+		Data: make(map[interface{}]interface{}),
 	}
 	for _, hook := range db.queryEventHooks {
 		hook.BeforeQuery(event)
--- a/hook.go
+++ b/hook.go
@@ -2,8 +2,6 @@ package pg
 
 import (
 	"fmt"
-	"runtime"
-	"strings"
 	"time"
 
 	"github.com/go-pg/pg/orm"
@@ -85,13 +83,7 @@ func (db *DB) queryStarted(
 		return nil
 	}
 
-	funcName, file, line := fileLine(2)
 	event := &QueryEvent{
-		StartTime: time.Now(),
-		Func: funcName,
-		File: file,
-		Line: line,
-
 		DB:      ormDB,
 		Query:   query,
 		Params:  params,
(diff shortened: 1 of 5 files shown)


vmihailenco:
Hi,

Thanks for the PR and sorry for the delay.

I would expect that query started and query processed hooks receive the same event and you can save some data in query started hook event that is available in query processed hook. Also it probably makes sense to rename hooks like this:

```
type Hooks interface {
    OnBeforeQuery(*QueryEvent)
    OnAfterQuery(*QueryEvent)
}

type QueryEvent struct {
    // ... old fields here and
    Data map[interface{}]interface{} // here you can save some data
}

db.AddHook(&structImplementingHooksInterface{})
```

What do you think?

edevil:
@vmihailenco That seems perfect for me, but that would be a breaking change. Is that Ok?

vmihailenco:
Yes, I think that is fine as long as we push new major version.

edevil:
@vmihailenco done, PTAL

edevil:
@vmihailenco ping :)

hook.go
```diff
@@ -62,43 +69,58 @@ func queryString(query interface{}) ([]byte, error) {
...
-	event := &QueryProcessedEvent{
-		StartTime: start,
-		Func:      funcName,
-		File:      file,
-		Line:      line,
+	event := &QueryEvent{
+		StartTime: time.Now(),
+		Func: funcName,
+		File: file,
+		Line: line,
```
vmihailenco: Can you please remove these params? I think it is reasonable to expect that users should save info they need in `Data` map.

vmihailenco:
I like it! One small change and this should be good to go.

hook.go
edevil: Done.

edevil:
@vmihailenco Implemented the change you suggested. PTAL :)

hook.go
```diff
@@ -62,84 +67,51 @@ func queryString(query interface{}) ([]byte, error) {
...
 }
 
-type queryProcessedHook func(*QueryProcessedEvent)
-
-// OnQueryProcessed calls the fn with QueryProcessedEvent
+// OnQueryEvent calls the interface with QueryEvent
 // when query is processed.
-func (db *DB) OnQueryProcessed(fn func(*QueryProcessedEvent)) {
-	db.queryProcessedHooks = append(db.queryProcessedHooks, fn)
+func (db *DB) OnQueryEvent(hook Hook) {
```
vmihailenco: I will change this myself, but I think this should be `AddHook` or `AddQueryHook`. I expect `OnSomething` to receive a function.

vmihailenco:
Thanks! Looks pretty good. Let me know if you spot that anything is still missing.

hook.go
edevil: Yeah, I didn’t give the name much thought, I agree.

edevil:
@MrBadge You can store `time.Now()` on the `Data` map in "BeforeQuery", and use it on "AfterQuery".

MrBadge:
> @MrBadge You can store `time.Now()` on the `Data` map in "BeforeQuery", and use it on "AfterQuery".

Thanks, that's indeed more convenient

PR #1110, opened 2019-02-10
Added support for tablespaces in models.

elliotcourant:
This will allow a `tableSpace` string field to be added to a struct and will take the value of the name in the SQL tag and set the table space when creating a table.

https://www.postgresql.org/docs/current/manage-ag-tablespaces.html

I've made it so that it will also quote the tablespace names just to be safe (to be consistent with quoting table names).

I would love feedback!

### DIFF
--- a/orm/table.go
+++ b/orm/table.go
@@ -52,6 +52,8 @@ type Table struct {
 	FullName           types.Q
 	FullNameForSelects types.Q
 
+	Tablespace types.Q
+
 	allFields     []*Field // read only
 	skippedFields []*Field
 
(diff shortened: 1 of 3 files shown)


vmihailenco:
Can this be an option on `tableName` like this:

```go
type CreateTableWithTablespace struct {
	tableName struct{} `sql:"tablespace:ssd"`

	String string
}
```

?

elliotcourant:
I like that approach much better, I'll make the adjustments and add my logic to the handling for the `tableName` field.

elliotcourant:
I've made the adjustment you recommended. Thank you for the feedback and the excellent library!

vmihailenco:
Thanks - I've simplified code a bit in the master branch.

Issue #1074, opened 2018-12-26
How do I handle the money data type in a model?


zignd:
There's a table with a column whose data type is money, I'd like to create a model capable of supporting this column. My initial attempt was to use a float64, but the library fails returning the following error `strconv.ParseFloat: parsing "$123.12": invalid syntax`. What is the recommended Go type to store values coming from money columns?

```go
type ProductPrice struct {
	ID        int
	ProductID int
	Product   *Product
	Price     float64
	Current   bool
	CreatedAt time.Time
	UpdatedAt *time.Time
}
```

Thanks in advance! :)

aligfl:
I have the same problem, any ideas on a workaround?

vmihailenco:
go-pg supports sql.Scanner and sql/driver.Valuer interfaces so if there are libs that work with database/sql - use them.

You can also use `ScanColumn` and `AppendValue` to implement something yourself - https://github.com/go-pg/pg/blob/master/pg.go#L116-L138

elliotcourant:
I have started working on a resolution for this. From what I'm finding there aren't many client libraries that necessarily support money types; so there isn't much to go off of. At the moment my fix only works for reading values but if there is still need/interest for this I will continue to develop it.

zignd:
@elliotcourant that's really useful! Please continue the work! And thanks for taking the time to implement that as addition to go-pg. :+1:

elliotcourant:
What locales are you using? Do they ever change, or is it always the same locale? Does the locale match the system locale for the machine running the go-pg/pg?

vmihailenco:
FYI [Don't use money](https://wiki.postgresql.org/wiki/Don%27t_Do_This)

PR #1111, opened 2019-02-11
Added some basic handling for money types.

elliotcourant:
When a dollar sign is present when parsing a float value the dollar sign
will now be removed. Previously it was causing an error to be returned
from strconv.ParseFloat.

This should resolve: https://github.com/go-pg/pg/issues/1074


vmihailenco:
This only works for '$' - what should we do for all the other currencies?

elliotcourant:
It looks like the currency symbol used is based on `lc_monetary`, I'll do some research and see if there is an easy way to handle that. 

I could just make a map of the currency symbols and have those symbols removed, but that seems very inefficient.

elliotcourant:
It might be more difficult than simply replacing some characters.
The following query is very simple and returns a result that could be easily parsed.

> ```sql
> SET lc_monetary TO 'EN_ie';
> SELECT 5.5::money;
> ```
> ```
> €5.50
> ```

But if we are using Russian currency it gets a lot more complicated.
> ```sql
> SET lc_monetary TO 'RU_ru';
> SELECT 5.5::money;
> ```
> ```
> 5,50 руб.
> ```

I think the best approach would be to detect the type returned from PostgreSQL using the OID and then if it is money; parse the value to a float using Regex. That way there is only overhead when actually using money types.

elliotcourant:
So after some more research, I've found a way to get `lc_monetary` information in Go by calling some C functions with cgo. Using `locale.h` should work on all platforms as well. I believe I can build something to parse anything returned from PostgreSQL in any format, but it would need to know the locale that Postgres is using. 

I will build out a function to do the parsing and post it here for you to take a look at, if you think that it will work and not cause problems then I can implement it into your library.

ORBAT:
As a go-pg user, I'm not really thrilled about extra the cgo dependency just to parse things related to monetary types, plus I'd assume it'd make maintenance of the package itself harder. 

Have you taken a look at the packages under https://godoc.org/golang.org/x/text, such as https://godoc.org/golang.org/x/text/language and https://godoc.org/golang.org/x/text/currency? It seems like they could help here, in combination with eg. https://github.com/cloudfoundry-attic/jibber_jabber to get the system locale in a platform-independent, pure Go way

ECourant:
I will look into those and see if I can make a pure go version. I agree a cgo dependency is a bit extreme. Essentially I just need the information from `lc_monetary`, if there is a pure go way to get that then that would be 110% ideal.

elliotcourant:
@ORBAT After doing some more research on the resources you mentioned and some other documentation I was able to dig up. I think I figured out a way. I'm going to make a separate (pure-go) package specifically for handling/parsing monetary values without cgo. Then I'll build a wrapper for that into go-pg/pg, this way it will be easier to maintain and the money parsing can be used as a standalone library.

elliotcourant:
Quick update.

I've finished creating a pure-go library for parsing and formatting monetary values. I will be redoing all my work in this PR over the next few days to get this implemented. Current the package supports ~200 different locales.

https://github.com/elliotcourant/gomonetary

elliotcourant:
I'm gonna close this MR for now, I haven't heard from the guys who wanted it. And I'm running into some complexities. I'll reopen if it becomes relevant again.

Issue #1141, opened 2019-03-19
Data race on `orm.Table.FieldsMap`


jayschwa:
I will post a test case if I can devise one. We are experiencing this issue after upgrading to v8.0.2 from v6.15.0.

Read: https://github.com/go-pg/pg/blob/v8.0.2/orm/table.go#L944
Write: https://github.com/go-pg/pg/blob/v8.0.2/orm/table.go#L690
```
WARNING: DATA RACE
Read at 0x00c00031e090 by goroutine 37:
  runtime.mapaccess1_faststr()
      /usr/local/go/src/runtime/map_faststr.go:12 +0x0
  github.com/go-pg/pg/orm.foreignKeys()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/table.go:944 +0x1f4
  github.com/go-pg/pg/orm.(*Table).tryRelationSlice()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/table.go:624 +0x14e0
  github.com/go-pg/pg/orm.(*Table).tryRelation()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/table.go:516 +0x10b
  github.com/go-pg/pg/orm.(*Table).initRelations()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/table.go:491 +0xde
  github.com/go-pg/pg/orm.(*Table).init()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/table.go:484 +0x38
  github.com/go-pg/pg/orm.(*tables).get()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/tables.go:86 +0x43b
  github.com/go-pg/pg/orm.(*tables).Get()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/tables.go:98 +0x51
  github.com/go-pg/pg/orm.GetTable()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/tables.go:18 +0x5e
  github.com/go-pg/pg/orm.newSliceTableModel()
      /go/pkg/mod/github.com/go-pg/pg@v8.0.2+incompatible/orm/model_table_slice.go:24 +0x46
  github.com/go-pg/pg/orm.newTableModelValue()
[...]

PR #1142, opened 2019-03-21
orm: fix race on FieldsMap

vmihailenco:
Fixes https://github.com/go-pg/pg/issues/1141


jayschwa:
I've been doing some testing with this branch and it seems promising. I have not seen any data races yet.

vmihailenco:
Thanks for testing

PR #1156, opened 2019-04-01
Allow for has many queries to be executed in a single round trip.

elliotcourant:
Currently if a model has a `has many` relationship and it is queried, then two SELECTs will be sent to PostgreSQL. 

Documented here: https://godoc.org/github.com/go-pg/pg#example-DB-Model-HasMany
Select user and all his active profiles with following queries:

### Current behavior:
```sql
SELECT "user".* FROM "users" AS "user" ORDER BY "user"."id" LIMIT 1;

SELECT "profile".* FROM "profiles" AS "profile"
WHERE (active IS TRUE) AND (("profile"."user_id") IN ((1)))
```

### Updated behavior:
```sql
SELECT
  "user"."id",
  "user"."name",
  json_agg(json_build_object('Id', "profile"."id", 'Lang', "profile"."lang", 'Active', "profile"."active", 'UserId', "profile"."user_id")) AS "profiles"
FROM "users" AS "user"
LEFT JOIN "profiles" AS "profile" ON "users"."id" = "profile"."user_id"
WHERE "profiles"."active" IS TRUE
```

This reduces the number of round trips needed to query data and has the potential to make it significantly easier to filter based on relational data.

This probably needs some work though, let me know what your thoughts are.

### DIFF
--- a/orm/join.go
+++ b/orm/join.go
@@ -228,6 +228,36 @@ func (j *join) appendHasOneColumns(b []byte) []byte {
 	return b
 }
 
+func (j *join) appendHasManyColumns(b []byte) []byte {
+	b = append(b, "json_agg(json_build_object("...)
+
+	for i, f := range j.JoinModel.Table().Fields {
+		jsonName, ok := f.Field.Tag.Lookup("json")
+		if jsonName == "-" {
+			// If the user is ignoring this json field then we cannot marshal it using this.
+			continue
+		}
+		if !ok || jsonName == "omitempty" {
+			// If a json field name isn't specified, default to the name of the go field.
+			jsonName = f.GoName
+		}
+
+		if i > 0 {
+			b = append(b, ", "...)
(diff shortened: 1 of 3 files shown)


elliotcourant:
Looks like it broke some things, but the concept still is there. @vmihailenco what are your thoughts?

vmihailenco:
I am skeptical for multiple reasons
- I personally don't use ORM capabilities (has-one, has-many etc) and I am not sure it is that useful to spend more time on it
- your query most likely does not work as is, e.g. I suspect it still returns duplicated rows for each left-joined row
- using JSON means we now have 2 de-serialization protocols which potentially introduces a lot of bugs/inconsistencies
- complex queries = complex plans = bad performance

I doubt I will be working on this any time soon...

elliotcourant:
Thats reason enough for me 👍

PR #1161, opened 2019-04-04
db: fix cancel request race condition on context timeout/cancellation

taiyangc:
When passing a timeout context into the db queries, there currently
exists a race condition where a select between query finish and context
done can choose the context, and go on to send a cancel request on a
conn that has already been freed. So a concurrent request may pick up
the same conn but is cancelled by this orphaned request.

There is a similar fix in https://github.com/lib/pq/pull/578 where we
use the same channel to pass flow between fn finish and cancel
goroutines. close() is async so a select cannot guarantee executing one
case at a time.

Added a test in race_test to prove the problem and the fix.

### DIFF
--- a/base.go
+++ b/base.go
@@ -130,16 +130,21 @@ func (db *baseDB) withConn(c context.Context, fn func(cn *pool.Conn) error) erro
 		fnDone = make(chan struct{})
 		go func() {
 			select {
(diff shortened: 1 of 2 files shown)


vmihailenco:
Looks good - thanks

Issue #1169, opened 2019-04-08
No automatic reconnect when using transactions


moogle19:
Hi, 
I got the following problem using go-pg:
When using normal Insert, Select... methods the reconnecting in case of a db restart is handled as it should. 

But when I use a transaction (with `db.Begin` and `tx.Insert`) I get a `write tcp [::1]:52557->[::1]:5432: write: broken pipe` error (instead of `dial tcp [::1]:5432: connect: connection refused` for an Insert without a transaction) and after the restart of the db is done, no reconnection takes place.

Tested with: 
go-pg: `v8.0.3` & `v8.0.0` & `v7.1.7`
Postgres: `10.7`

Sample code: 
```go
package main

import (
	"log"
	"time"

	"github.com/go-pg/pg"
)

type test struct {
	Test string
}

func main() {
	db := pg.Connect(&pg.Options{
		User:     "postgres",
		Database: "testing",
	})

	test := test{
		Test: "hello",
	}

	for {
		time.Sleep(time.Second)
		tx, err := db.Begin()
		if err != nil {
			log.Printf("failed to begin: %v", err)
			continue
		}
		err = tx.Insert(&test)
		if err != nil {
			continue
			log.Printf("failed to insert: %v", err)
		}
		err = tx.Commit()
		if err != nil {
			log.Printf("failed to commit: %v", err)
		}
	}
}
```

vmihailenco:
This is expected - you need to re-run transaction on error manually (if that is what you want). Otherwise go-pg could screw data in the database.

moogle19:
@vmihailenco 
That's what I am trying, but the `broken pipe` error is returned from `tx.Begin`, so I can't start a new transaction.
If I add a simple `db.Exec("SELECT 1")` before retrying to begin a transaction, it fixes the problem.

moogle19:
@vmihailenco 
I think the behaviour is cause by this line:
https://github.com/go-pg/pg/blob/2025d4012ba655acfa7430308d87fb2ea2f25bb0/internal/pool/pool_single.go#L90
The channel is never read again if the connection was terminated by a db restart or connection issue so the defer function which calls `p.pool.Remove` is never called, because the channel write blocks forever.

vmihailenco:
`tx.Begin` does not start new transaction. It is dummy function that just returns current transaction - https://github.com/go-pg/pg/blob/master/tx.go#L59-L62.

moogle19:
@vmihailenco 
[...]

PR #1170, opened 2019-04-09
Fix connection removal for single pool

moogle19:
`Remove` on the `SingleConnPool` now removes the connection
also from the source pool and resets the state of the
`SingleConnPool` to `Inited`.

This fixes an error that when a connection fails, it was readded to the pool and the next time a SingleConnPool is initiated, it would use the failed connection, so it would never be removed from the pool.

The `Remove` function is only called from `base.go` when `initConn` fails or when `freeConn` is called with a connection error, so I think it's save to always remove the connection also from the source pool, because there is (currently) no case when `SingleConnPool.Remove` is called on a healthy connection.

Fixes the problem described in https://github.com/go-pg/pg/issues/1169#issue-430352506

### DIFF
--- a/internal/pool/pool_single.go
+++ b/internal/pool/pool_single.go
@@ -82,12 +82,10 @@ func (p *SingleConnPool) Put(cn *Conn) {
 }
 
 func (p *SingleConnPool) Remove(cn *Conn) {
-	defer func() {
-		if recover() != nil {
-			p.pool.Remove(cn)
-		}
-	}()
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks for digging into this, but the problem here is that if tx fails in middle of something important this change will reset tx state. I think https://github.com/go-pg/pg/pull/1175 is more correct fix for the problem you revealed.

PR #1227, opened 2019-05-28
Pass and return context in hooks

vmihailenco:
Fixes https://github.com/go-pg/pg/issues/1226


axw:
Hi @vmihailenco. Since this is a breaking change - just wanted to check, will you be doing a major version bump on the next release?

vmihailenco:
Yep, v9 it will be

axw:
Thank you!

PR #1234, opened 2019-06-03
added: support for negated queries.

lilezek:
A proposal for adding negated queries. We found them useful to search for the opposite of already written queries.

### DIFF
--- a/orm/query.go
+++ b/orm/query.go
@@ -381,9 +381,27 @@ func (q *Query) Where(condition string, params ...interface{}) *Query {
 	return q
 }
 
+func (q *Query) WhereNot(condition string, params ...interface{}) *Query {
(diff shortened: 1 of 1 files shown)


vmihailenco:
I don't have objections against `WhereNotGroup` and `WhereOrNotGroup`, but `WhereNot` can be replaced with `Where("NOT "+condition, params...)` which is simple enough for such rare case. Could you remove `WhereNot` and `WhereOrNot`?

lilezek:
Maybe I'm missing something, but with the following code

```
q.Where("NOT 1 == 2")
```

I get the following fragment:

```
AND (NOT 1 == 2)
```

I think this fragment is not valid. Any idea?

vmihailenco:
Thank you

PR #1243, opened 2019-06-16
Initialize event.Stash

aashu16:


vmihailenco:
Please initialize it in before hook when you need it.

aashu16:
Thanks for clarifying. This breaks [genna](https://github.com/dizzyfool/genna) so I have created a pull request there instead.

For curiosity's sake, what was the reason for this change?

vmihailenco:
`make(map[interface{}]interface{})` is an allocation which is not needed in most cases now where there is `StartTime` field.

PR #1249, opened 2019-06-22
Use QuoteTableName on safe names

vmihailenco:


eparziale:
@vmihailenco After these changes, how do we now escape reserved words if they exist as column names?

i.e. Before these changes, the following model definition worked fine in go pg:

```
type Email struct {
	Id   int64
	From string `sql:"from"`
	To string
	Cc string
	Subject string
}

```
However, now the above model definition no longer works because the SQL is built WITHOUT double quotes around each column name.


i.e. the following create statement worked before (notice double quotes around reserved words **from** and **to**

```
CREATE TABLE IF NOT EXISTS emails 
(
	id bigserial, 
	"from" text, 
	"to" text, 
	cc text, 
	subject text
)
```

But now the SQL is built without quoting **from** and **to**

```
CREATE TABLE IF NOT EXISTS emails 
(
	id bigserial, 
	from text, 
	to text, 
	cc text, 
	subject text
)
```

Can you please advise on the best way to adjust our code to play nicely with the latest go-pg changes?

I tried overriding default column name following model def here: https://github.com/go-pg/pg/wiki/Model-Definition

But that didn't create double quotes.

Thank you,

Issue #1213, opened 2019-05-17
A missing query hook?


johnwiichang:
I found that we can make some complex types available with a missing query hook: `BeforeQuery(context.Context, orm.Query)`

I have tried to use range type (`tsrange`) in my project, and I found this type cannot work with `go-pg` lib directly しかし I can trans related fields with some tricks.

The type is defined like:

```go
type (
	A struct {
		Period Period `json:"-"`
	}
	
	Period struct {
		Start *time.Time
		Stop  *time.Time
	}
)

func (p Period) String() string {
  //trans two time_ref to "[from,to)"
}
```

At first, the go-pg will identify `Period` as a `jsonb`, but I just wanna `tsrange` instead. So I try to make a shadow struct:

```go
type aShadow struct {
	Period text `json:"-", sql:"period"`
}

type A struct {
	Period Period `json:"period", sql:"-"` //ignored by go-pg
	aShadow //the field named period will be noticed by our lib
}

func (a *A) prepare() {
  a.aShadow.Period = a.Period.String()
}
```

I use hooks to handle the `Insert`, `Update` actions, but the code is repeated without any extra operations required. The operation is boring and meaningless, so if there was `BeforeQuery` existed, I can reduce these repeated code.

I have forked a copy of code from master and implemented this function already. But I don't think it's a hard work to add this extra function... Would you like to give me some suggestions? or just tell me why `After` only without `Before`?

Have a nice coding day. (´･ω･`)

PR #1256, opened 2019-06-26
Add BeforeQuery Hook #1213

johnwiichang:
Some attempts were made for Issue #1213 to verify the test code and fix some potential issues after a while.


vmihailenco:
Hi,

Today I've changed existing model hooks and specifically removed `AfterQuery` hook - see https://github.com/go-pg/pg/wiki/Model-Hooks. The rationale is is that
- we already have more specific model hooks like BeforeInsert/Update/Delete etc
- there are db hooks with same names in https://github.com/go-pg/pg/blob/master/hook.go#L32-L33

Overall it is not clear how db and model hooks should play together. Should both of them be called? Which one is called first? Which one should be used when?

Since I don't have an answer to those questions it seems easier to just remove duplication.

johnwiichang:
In fact, I think it is not a good way to delete it directly. BeforeQuery is a hook that helps someone make some adjustments to the parameters and solves some of the unsupported classes mentioned in issue #1261 .
I think `BeforeInsert`, `BeforeUpdate` and `BeforeDelete` will let developers create a lot of boring duplicate code, and in most cases we may not need such an accurate operation. But I still understand that the basic library should adapt to many situations.

I will close this pr and keep these code in my fork. I need this feature to reduce some meaningless. It’s too boring to write so many duplicated code.

frederikhors:
I think we need both generic `BeforeQuery` and `AfterQuery`.

PR #1263, opened 2019-06-27
Added support for creating model tables with partition options.

elliotcourant:
This will allow a model's table to be created with options for partitions.
Allowing users to then create the individual partition tables manually
but still letting go-pg/pg manage the initial table creation.

The main reason for this is; we are using an account_id column for all of our tables. But we want to partition our tables using PostgreSQL's partition feature by account_id and hash. We also want to add some partitioning even further on some of those tables where we keep archived or deleted records in their own partition on a tablespace that resides on a slower hard drive. While keeping active records on a different partition on a faster tablespace.

This will make it so the initial creation of our tables is much easier, and we can manage creating the individual partitions manually by just inheriting the parent table.

More documentation here: https://www.postgresql.org/docs/12/ddl-partitioning.html

### DIFF
--- a/orm/table.go
+++ b/orm/table.go
@@ -54,6 +54,8 @@ type Table struct {
 
 	Tablespace types.Q
 
+	PartitionBy string
+
 	allFields     []*Field // read only
 	skippedFields []*Field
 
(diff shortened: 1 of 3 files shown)


vmihailenco:
I think it should be either

```
type CreateTableWithRangePartition struct {
	tableName struct{} `sql:"partitionBy:RANGE(time)"`

	Time   time.Time
	String string
}
```

or even

```
type CreateTableWithRangePartition struct {
	pgPartitionType struct{} `sql:"RANGE(time)"`

	Time   time.Time
	String string
}
```

Later scales better but perhaps we don't need it. Separating definition into partitionType and partitionBy only makes it harder to read...

elliotcourant:
I prefer the first one you mentioned, it seems very intuitive. I'll make those changes!

elliotcourant:
Should any escaping be done of the value of `partitionBy`? Or should that just be appended to the query and any resulting syntax error would just produce an error when the query is executed anyway?

vmihailenco:
I think no escaping is required - all is appended to the query as provided by the user. This way you can partition by expression or whatever.

elliotcourant:
Just made those changes 👍

PR #1270, opened 2019-07-02
Replace busy loops with sync.Once in SingleConnPool

JelteF:
Inspecting the code I saw some busy loops to do synchronized setup. Using sync.Once seems like a better solution for this.

### DIFF
--- a/internal/pool/pool_single.go
+++ b/internal/pool/pool_single.go
@@ -51,11 +51,10 @@ func (p *SingleConnPool) Clone() *SingleConnPool {
 }
 
 func (p *SingleConnPool) SetConn(cn *Conn) {
-	if atomic.CompareAndSwapUint32(&p.state, stateDefault, stateInited) {
+	p.setup().once.Do(func() {
+		atomic.StoreUint32(&p.state, stateInited)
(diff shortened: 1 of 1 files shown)


vmihailenco:
That loop should be executed 1-2 times - it is not particularly busy/hot. I don't think there is practical difference between this version and yours. But if you have a benchmark that say otherwise - please show it.

The problem with this version is that now pool can be inited after it was closed.

JelteF:
The main practical difference is that `Get` will keep returning an error instead of deadlocking on `Get` after the first `Get` failed. Which seems more desirable behaviour.

The pool cannot be used after it was closed because of this:
https://github.com/go-pg/pg/pull/1270/files#diff-c624debe44509052bc32240703942bebR142

vmihailenco:
@JelteF thanks for catching the deadlock.

After some consideration I think that this code is unnecessary complex. It is not called in a hot loop / concurrently and there is no reason to over-complicate things to save 100 nanoseconds. I've sent https://github.com/go-pg/pg/pull/1280 that uses good old `sync.Mutex` and hopefully make code more obviously correct.

But thanks for working on this!

PR #1325, opened 2019-08-03
Added support for omitting zero values from batch updates.

elliotcourant:
In a batch update it already correctly asserts that a zero value field
should be represented as such. But because it sends all data in the
update during a batch update it can still overwrite the existing value.

This solves that problem by adding a COALESCE around the value set
portion of the update statement so that if the value provided from the
ORM is null then it will not overwrite the existing database value.

This only affects updates done with slices. Single record updates 
behave as they should currently.

### DIFF
--- a/orm/update.go
+++ b/orm/update.go
@@ -229,8 +229,16 @@ func (q *updateQuery) appendSetSlice(b []byte) ([]byte, error) {
 
 		b = append(b, f.Column...)
 		b = append(b, " = "...)
+		if q.omitZero {
+			b = append(b, "COALESCE("...)
+		}
 		b = append(b, "_data."...)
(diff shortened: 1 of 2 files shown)


vmihailenco:
Nice idea - thanks!

elliotcourant:
I found an issue with this just now running unit tests on our rest-api. 
Some columns on some types can cause the following error: `ERROR: COALESCE types text and bigint cannot be matched` because in the from clause I'm not casting some null values, and without any type information PostgreSQL assumes it is text. So if I provide a null value for a bigint column with no type cast in the from clause it will cause this error. 

I will submit a pull request to fix this shortly.

elliotcourant:
This should be fixed in #1329

PR #1327, opened 2019-08-03
Don't accept relantion in Query.Column

vmihailenco:


eparziale:
@vmihailenco  Can you please provide some background on the reasoning behind this change? 

Was there a team using go-pg that had an issue with the previous implementation where both columns and relations could be passed as columns? Was there a very good reason to make this change in how relations are selected?

We were able to adjust our application code to handle these changes, but this was a significantly breaking change that took some thought and careful code changes to play nicely with an established code base using go-pg.

I just want to understand whether there was a large value-add in making this change.

Also, some of the FAQ docs still show the old relation query method:
https://github.com/go-pg/pg/wiki/FAQ

See:
`db.Model(&books).Column("book.*", "Author").Limit(100).Select()`

Thanks!

E

vmihailenco:
The main reason is that `Relation` returns an error when there is no relation with such name. That somewhat helps new users which assume that since there are no errors all is fine in their code.

Also `Relation` is more explicit (which makes it special / stand out) and accepts a 2nd argument which is often useful too.

Overall the majority of questions I receive are about ORM so I am trying to do something about it. Ideally I should probably start using some deprecation policy but that is time consuming...

eparziale:
Ok understood. That makes sense.

Thanks very much for the additional context and details.

Change is sensible from my perspective.

PR #1329, opened 2019-08-03
Added handling of serial column types in updates.

elliotcourant:
When performing a bulk update previously it would try to cast the fields
that were "bigserial" or "serial" to those types in the from clause. But
because they are not actual types in PostgreSQL and are just aliases, it
would cause an error when executing the query. This resolves that by
making sure that serial columns are cast to their base type instead.

This fixes the error:  `[42704] ERROR: type "bigserial" does not exist`

### DIFF
--- a/orm/update.go
+++ b/orm/update.go
@@ -222,6 +222,11 @@ func (q *updateQuery) appendSetSlice(b []byte) ([]byte, error) {
 		fields = q.q.model.Table().DataFields
 	}
 
+	var table *Table
+	if q.omitZero {
+		table = q.q.model.Table()
+	}
+
 	for i, f := range fields {
 		if i > 0 {
 			b = append(b, ", "...)
@@ -229,13 +234,17 @@ func (q *updateQuery) appendSetSlice(b []byte) ([]byte, error) {
 
 		b = append(b, f.Column...)
 		b = append(b, " = "...)
-		if q.omitZero {
+		if q.omitZero && table != nil {
 			b = append(b, "COALESCE("...)
 		}
 		b = append(b, "_data."...)
 		b = append(b, f.Column...)
-		if q.omitZero {
+		if q.omitZero && table != nil {
 			b = append(b, ", "...)
+			if table.Alias != table.FullName {
+				b = append(b, table.Alias...)
+				b = append(b, '.')
+			}
 			b = append(b, f.Column...)
 			b = append(b, ")"...)
 		}
@@ -310,10 +319,8 @@ func (q *updateQuery) appendValues(
 		} else {
 			b = f.AppendValue(b, indirect(strct), 1)
 		}
-		if f.HasFlag(customTypeFlag) {
-			b = append(b, "::"...)
-			b = append(b, f.SQLType...)
-		}
+		b = append(b, "::"...)
+		b = append(b, f.SQLType...)
 	}
 	return b, nil
 }
--- a/orm/update_test.go
+++ b/orm/update_test.go
@@ -10,6 +10,11 @@ type UpdateTest struct {
 	Value string `sql:"type:mytype"`
 }
 
+type SerialUpdateTest struct {
+	Id    uint64 `sql:"type:bigint,pk"`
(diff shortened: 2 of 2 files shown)


elliotcourant:
This also resolves an issue with the COALESCE usage when updating non-zero values.

When you specify a null value without specifying a type PostgreSQL assumes the type to be text. But then if you try to COALESCE that null value against a non-text column an error will be returned.

This pull request also fixes that issue by always specifying a type in the from clause for all fields to eliminate any type confusion on the server side.

elliotcourant:
There is also still an issue with updating at the moment because of ambiguous column names.

elliotcourant:
This now also fixed the error: `ERROR #42702 column reference "value" is ambiguous`

By formatting the query like this:
```sql
UPDATE "update_tests" AS "update_test"
SET "value" = COALESCE(_data."value", "update_test"."value")
FROM (VALUES (1::BIGINT, 'hello'::mytype), (2::BIGINT, NULL::mytype)) AS _data("id", "value")
WHERE "update_test"."id" = _data."id"
```

Instead of like this:
```sql
UPDATE "update_tests" AS "update_test"
SET "value" = COALESCE(_data."value", "value")
FROM (VALUES (1::BIGINT, 'hello'::mytype), (2::BIGINT, NULL::mytype)) AS _data("id", "value")
WHERE "update_test"."id" = _data."id"
```

vmihailenco:
Could you please rebase this on https://github.com/go-pg/pg/pull/1331? With that PR go-pg should not produce `bigserial` type any more. And you should not use it when defining field type too. Instead define it as:

```go
type SerialUpdateTest struct {
	Id    uint64 `sql:"type:bigint,pk"`
	Value string
}
```

On latest version you should probably identity instead of serial - https://github.com/go-pg/pg/issues/1222.

orm/update.go
```diff
@@ -331,3 +335,16 @@ func appendWhereColumnAndColumn(b []byte, alias types.Q, fields []*Field) []byte
 	}
 	return b
 }
+
+func appendTypeCast(b []byte, f *Field) []byte {
```
vmihailenco: I think this can be removed with https://github.com/go-pg/pg/pull/1331

orm/update_test.go
```diff
@@ -10,6 +10,11 @@ type UpdateTest struct {
 	Value string `sql:"type:mytype"`
 }
 
+type SerialUpdateTest struct {
+	Id    uint64 `sql:"type:bigserial"`
```
vmihailenco: type:bigint,pk

orm/update.go
```diff
@@ -222,20 +222,27 @@ func (q *updateQuery) appendSetSlice(b []byte) ([]byte, error) {
...
-		if q.omitZero {
+		if q.omitZero && table != nil {
 			b = append(b, "COALESCE("...)
 		}
 		b = append(b, "_data."...)
 		b = append(b, f.Column...)
-		if q.omitZero {
+		if q.omitZero && table != nil {
 			b = append(b, ", "...)
+			b = append(b, table.Alias...)
```
vmihailenco: Can you please use:

```
if table.Alias != table.FullName {
    b = append(b, table.Alias...)
    b = append(b, '.')
}
```

Because on some Postgres version alias is not supported.

orm/update.go
vmihailenco: I.e. replaced with

```
b = append(b, "::"...)
b = append(b, f.SQLType...)
```

vmihailenco:
Overall this looks good, but let's try to clean up the code a bit...

elliotcourant:
Thanks @vmihailenco I'll make those changes 👍

vmihailenco:
Gentle ping

elliotcourant:
I've not forgotten about this, sorry. Just been distracted.

elliotcourant:
@vmihailenco should be good to go now. The build passed so I squashed all of my commits.

vmihailenco:
Thanks!

PR #1335, opened 2019-08-07
fix unbalance pool

### LINKED ISSUE #1334: ConnPool is unbalanced
ConnPool get a connection from the tail of the idleConns, and return the connection to the tail of the idleConns, which causes the connections at the end of the idleConns used heavily.

ruccsbingo:
fixed #1334

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -302,9 +302,8 @@ func (p *ConnPool) popIdle() *Conn {
 		return nil
 	}
 
-	idx := len(p.idleConns) - 1
-	cn := p.idleConns[idx]
-	p.idleConns = p.idleConns[:idx]
+	cn := p.idleConns[0]
+	p.idleConns = p.idleConns[1:]
 	p.idleConnsLen--
 	p.checkMinIdleConns()
 	return cn


internal/pool/pool.go
```diff
@@ -302,9 +302,8 @@ func (p *ConnPool) popIdle() *Conn {
 		return nil
 	}
 
-	idx := len(p.idleConns) - 1
-	cn := p.idleConns[idx]
-	p.idleConns = p.idleConns[:idx]
+	cn := p.idleConns[0]
+	p.idleConns = p.idleConns[1:]
```
vmihailenco: As I understand there is a memory leak here since first element of the slice is inaccessible but can not be freed.

lemonlinger: maybe below code could fix the memory leak ? @vmihailenco 
`
cn := p.idleConns[0]
p.idleConns[0] = nil
p.idleConns = p.idleConns[1:]
`

vmihailenco: What is the motivation? Why balancing connection usage is important?

lemonlinger: Because unbalanced connection usage makes few backends bear more traffics that will increase the average response time per query.

vmihailenco: Do you mean postgres backends or you are using some load balancer that is not able to re-distribute queries that come on the same connection?

lemonlinger: I mean postgres backends.

vmihailenco: Okay, thanks. Before making any changes we need a benchmark or something that proves that uneven queries distribution between Postgres backends affects performance.

lemonlinger: Actually we happen to encounter this issue in our production. I can provide some screenshots from grafana graphs showing how the response time affected by uneven queries. But it's non-trivial to give a minimal demo of code to reproduce this issue.

PR #1360, opened 2019-08-24
Allow for bulk deletes with custom where clause.

elliotcourant:
Previously if you were to add your own where clause onto a delete
statement. pg would ignore the original clause (where ID in...) and
would instead only use the clause you had provided. This change makes
it so pg will attempt to use both with an AND so that all conditions
must be met for a record to be deleted when deleting from an array of
models.

This will allow more security in a situation where you are restricting
what a single account could do, as well as other use cases.

### DIFF
--- a/orm/delete.go
+++ b/orm/delete.go
@@ -82,19 +82,20 @@ func (q *deleteQuery) AppendQuery(fmter QueryFormatter, b []byte) (_ []byte, err
 	b = append(b, " WHERE "...)
 	value := q.q.model.Value()
 	if q.q.isSliceModelWithData() {
+		table := q.q.model.Table()
+		err = table.checkPKs()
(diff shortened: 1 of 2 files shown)


vmihailenco:
That is what we had in https://github.com/go-pg/pg/commit/8bc539e21436bf29621fcbc98b1a9795b80c9648#diff-56fa26fc2a8364369e2ab7ee66803637.

The problem with that approach is that when slice is empty the where clause by pk is completely omitted and the result is poorly predictable. E.g. `WHERE t.id IN (1, 2, 3) AND something` vs `WHERE something`.

It is safer to use `q.WherePK().Where("something").Delete()`.

elliotcourant:
That makes total sense a d is much easier!

Issue #1388, opened 2019-09-20
v9 appears to have broken type override in pk


bennichols:
Sorry if I've missed this elsewhere, but in migrating code to v9, my structs now throw errors when trying to generate tables.

This code used to work correctly, creating. bigint column using the default id_generator function:
```
type DiagnosticsErrorLog struct {
	ID        	int64     		`sql:"type:bigint,default:id_generator(),pk" json:"id,string"`
}
```

The SQL it generates is now this, trying to use the bigserial type instead, which doesn't work with the function:
```
CREATE TABLE "diagnostics_error_logs" ("id" bigserial DEFAULT id_generator(), PRIMARY KEY ("id"))
```

Is there something else I'm missing in the migration? Thanks!

PR #1391, opened 2019-09-20
orm: always use SQL type provided by the user

vmihailenco:
Fixes https://github.com/go-pg/pg/issues/1388


vmihailenco:
@elliotcourant pinging you just in case. If you are using `sql:",pk,type:bigint"` you will need to update it to be  `sql:",pk,type:bigserial"`. go-pg will normalize the type to "bigint" to use it in batch updates.

PR #1396, opened 2019-09-22
Added support for multiple named unique constraints for a single col.

elliotcourant:
This makes it so that a single column can be used in multiple unique
constraints by wrapping the unique name with single quotes and comma
separating the names of the unique constraints. But will still support
existing unique constraint formats.

The use case for this was we were trying to keep track of orders from multiple marketplaces (Amazon, eBay, etc...) and we wanted our internal order number to be unique per the tenant who owns the orders. But we wanted to make sure that within that each remote order ID was unique within a marketplace, and that the number was unique within that marketplace. But doing that within the same constraint would allow the same ID and number to appear multiple times in different combinations within a marketplace.

This just adds support for that feature as well as a basic test just to verify that it works.


orm/table.go
```diff
@@ -382,11 +382,15 @@ func (t *Table) newField(f reflect.StructField, index []int) *Field {
 	if v, ok := sqlTag.Options["unique"]; ok {
 		if v == "" {
 			field.SetFlag(UniqueFlag)
-		} else {
+		}
+		// Split the value by comma, this will allow multiple names to be specified.
+		// We can use this to create multiple named unique constraints where a single column
+		// might be included in multiple constraints.
+		for _, uniqueName := range strings.Split(strings.ReplaceAll(v, "'", ""), ",") {
```
vmihailenco: Please use `tagparser.Unquote(uniqueName)` to remove `'`.

orm/table_create_test.go
```diff
@@ -84,12 +84,19 @@ type CreateTableWithHashPartition struct {
 	String    string
 }
 
+type CreateTableWithMultipleNamedUniques struct {
+	ID               int    `sql:",pk"`
+	AccountID        int    `sql:",unique:'per_account,per_store'"`
```
vmihailenco: This looks good enough, but I am thinking that maybe it should be `sql:",unique:per_account,unique:per_store'"` and `sqlTag.Options` should be `map[string][]string` - i.e. support multiple values. Do you have any objections?

vmihailenco: Nevermind. I guess I am overthinking it. Please just use unquote and leave the rest as is.

elliotcourant:
Sorry for the late follow up on this. Did you just make those changes?

vmihailenco:
Yep. should be already done.

PR #1400, opened 2019-09-24
orm: deprecate sql tag

vmihailenco:


elliotcourant:
This is going to be a big change, are you going to make this v10? or is v9 still considered unstable and thus can have breaking changes?

vmihailenco:
It is deprecation - sql tags will continue working as expected. Let me know if they don't. They will be completely removed in v10.

PR #1426, opened 2019-10-11
Add go.mod to v8

dizzyfool:
Since there is a lot of breaking changes in v9 I suggest adding go mod versioning in last stable version (v8)


vmihailenco:
It is a breaking change and v8 does not receive any updates. So what is the point?

dizzyfool:
Checked on my repos and indeed it will break everything for non-go-mod projects

PR #1427, opened 2019-10-11
Loosen up error matching in test

yuzefovich:
Previously, the test that checks that an error is returned when
PREPARE'ing a bogus SQL statement would check for the precise error
message. Now, the test will check that the correct error code is
returned as well as a reasonable error message (i.e. the matching
is less strict yet should be sufficient). The reason for the change
is that other databases that "speak" Postgres (for example CockroachDB)
might return an error that slightly deviates from PG but is still
reasonable.


vmihailenco:
Thanks!

PR #1429, opened 2019-10-11
Introduce environment variables to customize running tests

yuzefovich:
This commit adds the support of some environment variables (namely,
PGHOST, PGPORT, PGUSER, PGDATABASE, and PGSSLMODE) that override
certain options when running tests. The inspiration is taken from
lib/pq tests (https://www.postgresql.org/docs/9.3/libpq-envars.html).
For every option, the corresponding environment variable is first
checked. If it is present, then the value is used; if it is not present
or empty, then we fall back to using the default values.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"math"
 	"net"
+	"os"
 	"strings"
 	"testing"
 	"time"
@@ -27,12 +28,7 @@ func TestGinkgo(t *testing.T) {
 
 func pgOptions() *pg.Options {
 	return &pg.Options{
-		User:     "postgres",
-		Database: "postgres",
-
-		TLSConfig: &tls.Config{
-			InsecureSkipVerify: true,
-		},
+		TLSConfig: getTLSConfig(),
 
 		MaxRetries:      1,
 		MinRetryBackoff: -1,
@@ -49,6 +45,16 @@ func pgOptions() *pg.Options {
 	}
 }
 
+func getTLSConfig() *tls.Config {
+	pgSSLMode := os.Getenv("PGSSLMODE")
+	if pgSSLMode == "disable" {
+		return nil
+	}
+	return &tls.Config{
+		InsecureSkipVerify: true,
+	}
+}
+
 var _testDB *pg.DB
 
 func testDB() *pg.DB {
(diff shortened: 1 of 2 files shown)


db_test.go
```diff
@@ -27,12 +28,10 @@ func TestGinkgo(t *testing.T) {
 
 func pgOptions() *pg.Options {
 	return &pg.Options{
-		User:     "postgres",
-		Database: "postgres",
+		User:     pg.GetPGUser(),
+		Database: pg.GetPGDatabase(),
```
vmihailenco: I think you can leave these empty so they are auto-populated later in options.go

options.go
```diff
@@ -142,6 +143,44 @@ func (opt *Options) init() {
...
+// GetPGDatabase returns the database name to be used when interacting with the
+// database server.
+func GetPGDatabase() string {
+	pgDatabase := os.Getenv("PGDATABASE")
+	if pgDatabase == "" {
+		pgDatabase = "postgres"
+	}
+	return pgDatabase
+}
+
```
vmihailenco: Can you please unexport all these functions? I don't think they should be part of the public API.

I think we don't need these functions at all - just inline the code in options init, e.g.

```
if opt.Database == "" {
    opt.Database = env("PGDATABASE", "postgres")
}

func env(key, defValue string) string {
}
```

options_test.go
```diff
@@ -132,7 +132,7 @@ func TestParseURL(t *testing.T) {
 		},
 		{
 			"postgres://vasya:pupkin@somewhere.at.amazonaws.com/postgres",
-			"somewhere.at.amazonaws.com:5432",
+			"somewhere.at.amazonaws.com:" + GetPGPort(),
```
elliotcourant: Given that these are not actually used for connections, and are purely for testing the parsing capability, i would not change any of these.

options.go
yuzefovich: Thanks, done.

options_test.go
yuzefovich: Thanks. I think I went overboard with the changes, reverted.

db_test.go
yuzefovich: Thanks, done.

vmihailenco:
Thank you!

PR #1449, opened 2019-11-02
Fixed bug when joining a non soft delete table with a soft delete table.

elliotcourant:
When a non soft delete table was the base table the query builder would
panic when it tried to add the soft delete criteria to the join ON
criteria. This was because we were properly evaluating that a soft
delete filter should be added on the join. But we were adding it for the
wrong table and trying to add it for a table where the soft delete
column was nil causing a panic.

This resolves the problem by making the ON filter for a soft delete join
for the actual join/soft deletable table rather than the base table.

To test this you can run the following test on the current master branch 9ab62c0f9c17851e2926c019d0262f198576930e and see how it fails:

```golang
var _ = Describe("SoftDeleteModel", func() {
	It("will join a non-SoftDelete with a SoftDelete", func() {
		type NonSoftDeleteModel struct {
			Id                int `pg:",pk"`
			Name              string
			SoftDeleteModelId int
			SoftDeleteModel   *SoftDeleteModel
		}

		type SoftDeleteModel struct {
			Id        int
			DeletedAt time.Time `pg:",soft_delete"`
		}

		q := NewQuery(nil, &NonSoftDeleteModel{}).Relation("SoftDeleteModel")

		s := selectQueryString(q)
		Expect(s).To(Equal(`SELECT "non_soft_delete_model"."id", "non_soft_delete_model"."name", "non_soft_delete_model"."soft_delete_model_id", "soft_delete_model"."id" AS "soft_delete_model__id", "soft_delete_model"."deleted_at" AS "soft_delete_model__deleted_at" FROM "non_soft_delete_models" AS "non_soft_delete_model" LEFT JOIN "soft_delete_models" AS "soft_delete_model" ON ("soft_delete_model"."id" = "non_soft_delete_model"."soft_delete_model_id") AND "soft_delete_model"."deleted_at" IS NULL`))
	})
})
```

### DIFF
--- a/orm/join.go
+++ b/orm/join.go
@@ -189,6 +189,17 @@ func (j *join) appendBaseAlias(b []byte) []byte {
 	return append(b, j.BaseModel.Table().Alias...)
 }
 
+func (j *join) appendSoftDelete(b []byte, flags queryFlag) []byte {
+	b = append(b, '.')
+	b = append(b, j.JoinModel.Table().SoftDeleteField.Column...)
+	if flags&deletedFlag != 0 {
+		b = append(b, " IS NOT NULL"...)
+	} else if flags&allWithDeletedFlag == 0 {
+		b = append(b, " IS NULL"...)
+	}
+	return b
+}
+
 func appendAlias(b []byte, j *join) []byte {
 	if j.hasParent() {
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks!

PR #1453, opened 2019-11-06
Recursive WITH

tpoxa:
Here is my attempt to make recursive with ASAP. May be useful for someone.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/go-pg/pg/v9
+module github.com/tpoxa/pg/v9
 
 go 1.11
 
--- a/orm/query.go
+++ b/orm/query.go
@@ -23,8 +23,9 @@ const (
 )
 
 type withQuery struct {
-	name  string
-	query QueryAppender
+	name      string
+	recursive bool
+	query     QueryAppender
 }
 
 type joinQuery struct {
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/go-pg/pg/v9
+module github.com/tpoxa/pg/v9
```
elliotcourant: Please don't change the package when submitting a pull request. Go mod should allow you to do any changes you need without needing to do this.

orm/query.go
```diff
@@ -219,29 +220,44 @@ func (q *Query) AllWithDeleted() *Query {
 
 // With adds subq as common table expression with the given name.
 func (q *Query) With(name string, subq *Query) *Query {
-	return q._with(name, newSelectQuery(subq))
+	return q._with(name, newSelectQuery(subq), false)
+}
+
+func (q *Query) WithRecursive(name string, subq *Query) *Query {
```
elliotcourant: Could you add just a very brief comment here like you have below in the wrap function.

orm/query.go
```diff
@@ -1484,6 +1500,9 @@ func (q *Query) appendReturning(fmter QueryFormatter, b []byte) (_ []byte, err e
 func (q *Query) appendWith(fmter QueryFormatter, b []byte) (_ []byte, err error) {
 	b = append(b, "WITH "...)
 	for i, with := range q.with {
+		if with.recursive {
+			b = append(b, "RECURSIVE "...)
```
elliotcourant: I think you should add something to check whether or not a space needs to be appended here. If `i` is greater than 0 then this will result in `WITH RECURSIVE , ` rather than `WITH RECURSIVE, `. It's not a huge deal but it makes the generated query look a lot better when debugging.

jh125486:
Is there an update to get this merged in for release?

PR #1479, opened 2019-12-14
Add "schema" parameter to "tableName" description tag.

### LINKED ISSUE #1468: Multiple schema support
Currently to make a query from table located in non-default (or currently selected schema) I use this hack
```go
type queryResult struct {
		tableName struct{} `sql:"information_schema.columns,alias:columns"`
                Name string `sql:"name"`
}
```
 which generates a correct query.
I think it will be better to have posibility to set schema name using a tag in tableName (like ```sql:"columns,schema:information_schema"```) and in query builder (like ```db.Model().SchemaExpr(schema)...```.

xakep666:
It allows to specify a schema name for particular model.
I.e. you can bravely list all columns in your database using
"information_schema" tables with such model:
```go
type Column struct {
  tableName struct{} `pg:"columns,schema:information_schema"`

  DBTableName string `pg:"table_name"`
  ColumnName  string `pg:"column_name"`
}
```
Should close #1468.

### DIFF
--- a/orm/table.go
+++ b/orm/table.go
@@ -62,6 +62,7 @@ type Table struct {
 	ModelName string
 
 	Name               string
+	Schema             types.Safe
 	FullName           types.Safe
 	FullNameForSelects types.Safe
 
@@ -145,8 +146,13 @@ func (t *Table) init2() {
 }
 
 func (t *Table) setName(name types.Safe) {
-	t.FullName = name
-	t.FullNameForSelects = name
+	if t.Schema != "" {
+		t.FullName = t.Schema + "." + name
+		t.FullNameForSelects = t.Schema + "." + name
+	} else {
+		t.FullName = name
+		t.FullNameForSelects = name
+	}
 	if t.Alias == "" {
 		t.Alias = name
 	}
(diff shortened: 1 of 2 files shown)


markuskobler:
@vmihailenco We would also be interested in schema functionally. Any chance of getting this PR reviewed?

vmihailenco:
As is I don't see how `pg:"some_name,schema:some_schema"` is any better than `pg:"some_schema.some_name"`. Theoretically it is possible to build some functionality on top of this PR later, e.g. adding some ability to change model schema. But again - it is not clear how something like `q.Schema("schema_name").Table("table_name")` is better than `q.Table("schema_name.table_name")`...

vmihailenco:
@xakep666 my apologies for such late feedback.

xakep666:
@vmihalienco I thought about `pg:"some_schema.some_name"`variant but currently it's treated as single table name  (dot escaped) and changing this behaviour breaks backwards compatibility. So I decided to add schema name as optional parameter in tag
Adding Schema option to builder looks interesting and probably I'll try to implement it here.

vmihailenco:
What I still don't understand is what we are trying to achieve here. If you just need to specify schema name - use `pg:"some_schema.some_name"` - that is what I am using for years.

>Adding Schema option to builder looks interesting and probably I'll try to implement it here.

Well, I actually think that `q.Table("schema_name.table_name")` is better - it is smaller API and less work required in go-pg. If `q.Schema("schema_name").Table("table_name")` is going to produce the same result as `q.Table("schema_name.table_name")` - then I don't see the point in adding it.

PR #1485, opened 2019-12-22
Do not use sql `DEFAULT` value when use_zero flag is on

noomz:
In my opinion, go-pg can use as an intermediate tool to the DB, but, it doesn't need to reflect the real DB structure and constraints.

So, if there was situation when model definition doesn't reflect DB column, use `DEFAULT` when `use_zero` and `default` tags use together can lead to `NULL` value in the database, even if I define `default` tag as empty value, go-pg should use my defined value as-is.

### DIFF
--- a/orm/insert.go
+++ b/orm/insert.go
@@ -185,6 +185,8 @@ func (q *insertQuery) appendValues(
 		switch {
 		case q.placeholder:
 			b = append(b, '?')
+		case !f.NullZero() && f.HasZeroValue(strct):
+			b = f.AppendValue(b, strct, 1)
 		case (f.Default != "" || f.NullZero()) && f.HasZeroValue(strct):
 			b = append(b, "DEFAULT"...)
 			q.addReturningField(f)
--- a/orm/insert_test.go
+++ b/orm/insert_test.go
@@ -162,6 +162,16 @@ var _ = Describe("Insert", func() {
 		Expect(s).To(Equal(`INSERT INTO "insert_default_tests" ("id", "value") VALUES (1, DEFAULT) RETURNING "value"`))
 	})
 
+	It("supports value when default value and use_zero is used together", func() {
+		type EmptyValueModel struct {
+			Value string `pg:",use_zero,default:''"`
+		}
+		q := NewQuery(nil, &EmptyValueModel{})
+
+		s := insertQueryString(q)
+		Expect(s).To(Equal(`INSERT INTO "empty_value_models" ("value") VALUES ('')`))
+	})
+
 	It("supports RETURNING NULL", func() {
 		q := NewQuery(nil, &InsertDefaultTest{Id: 1}).Returning("NULL")
 


vmihailenco:
`default:''` suggests that Postgres will generate `''` when it receives special `DEFAULT` value. So what is wrong with current behavior?

noomz:
> `default:''` suggests that Postgres will generate `''` when it receives special `DEFAULT` value. So what is wrong with current behavior?

With empty `Value`
```go
type EmptyModel struct {
  Value string `pg:"use_zero,default:''"
}
```
will generate `INSERT INTO "empty_model" VALUES (DEFAULT) RETURNING "value"` instead of `INSERT INTO "empty_model" VALUES ('')` which will be wrong (value will be null) if I didn't define default value as empty string in database, eg: `value varchar(100)` which is not `value varchar(100) default ''`.

vmihailenco:
>if I didn't define default value as empty string in database, eg: value varchar(100) which is not value varchar(100) default ''.

That is what you should do, because `default:''` means exactly that.

noomz:
> because `default:''` means exactly that.

From my perspective, `default:''` means empty string, it should be straightforward, not `DEFAULT`. I suggest that we should have another flag, like `default:_` or `default:DEFAULT` for such intention.

vmihailenco:
And how we should handle `default:'123'` or `default:now()`?

noomz:
That's why I proposed another flag (or any other ways). I think there are conflicts here:

1. `default:hello` has 2 meanings. First, it only indicates the **database** default value, not the GO struct value if omitted. Second, value is converted to `DEFAULT` if equal to tag value.
2. If I want to save an empty string, I need to provide the `use_zero` flag. Which **do nothing** to the database. It only a flag to test if the value should be NULL or empty (not null) to prepare a query.
3. `notnull` flag is used to mark `NOT NULL` to a column. Which do nothing to GO struct.

So the summary is:
1. not null : define the column.
2. use_zero : test the value. If a flag is not set. then transform.
3. default : define the column, test the value and then transform it.

It's so confusing that we need to concentrate that we are dealing with the database definition or the value to set.

We known that GO default value is the zero value of its type. We can't know if someone wants to save the zero value or default value if omitted. Without my patch I can't do this:

```go
type Model struct {
    S1 string `pg:",use_zero,default:hello"`
}
q := NewQuery(nil, &Model{})
q.Set("s1", "")
s := insertQueryString(q)
Expect(s).To(Equal(`INSERT INTO "models" ("s1") VALUES ('')`)) // will fail
Expect(s).To(Equal(`INSERT INTO "models" ("s1") VALUES (DEFAULT) RETURNING "s1"`)) // will pass, but not what I expected
q.Set("s1", nil) // also cannot do this
```

But with the patch:
```go
q.Set("s1", "")
s := insertQueryString(q)
Expect(s).To(Equal(`INSERT INTO "models" ("s1") VALUES ('')`)) // will pass
// but what if I want "hello" value
q.Set("s1", "hello")
q.Set("s1", nil) // and still, can't do this
```
Don't hesitate to tell what I'm missing. Or you can suggest the best practice to use `go-pg`.

vmihailenco:
The problem with `pg:",use_zero,default:hello"` generating `VALUES ('')` for zero value is that `default:hello` becomes useless - go-pg will never generate `VALUES(DEFAULT)` and Postgres will never assign default `hello` value for the cell.

So why do you need `default:hello` or `default:''`? What goal does it achieve?

>q.Set("s1", nil) // also cannot do this

`Set` is only for updates and it used as `q.Set("column = ?", "value")`. Did you try `q.Value("column", "''")`?

noomz:
OK, I can rest my case. Yes, you're right, I agree with your explanation and still can't find the right use case for it. For now, I still can use `go-pg` without that patch. Thank you 👍  (I will come back if I have any useful idea 😄 )

Issue #1496, opened 2020-01-08
soft delete with bigint ( unix epoch )


darioielardi:
Hello, thanks for the great work with this project.

I'm using the soft delete functionality, but I'd like to have my deleted_at column as a bigint ( set to unix epoch on deletion ), which is currently not allowed.
I can work on the implementation myself and make a PR, just let me know if you'd be willing to merge.

vmihailenco:
I will definitely take a look - whether it will be merged or not depends on how much code such feature will require...

darioielardi:
Should be a very simple and concise implementation, just a couple of if statements, queries handling doesn't need to be modified at all.
Seems like there are no tests for soft delete, is that correct?

vmihailenco:
There are some tests but definitely not full coverage. Just grep for `pg:",soft_delete"` - db_test.go:2192 is the main location.

PR #1497, opened 2020-01-08
soft delete with bigint column

darioielardi:
#1496

### DIFF
--- a/orm/model_table_struct.go
+++ b/orm/model_table_struct.go
@@ -2,6 +2,7 @@ package orm
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"reflect"
 	"strings"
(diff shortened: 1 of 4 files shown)


orm/model_table_struct.go
```diff
@@ -362,15 +363,29 @@ func (m *structTableModel) join(
 func (m *structTableModel) setSoftDeleteField() {
 	field := m.table.SoftDeleteField
 	value := field.Value(m.strct)
+	kind := value.Kind()
 
+	unixNow := time.Now().UnixNano() / 1e6
```
vmihailenco: This should be `/ time.Millisecond`. Overall - why do you use milliseconds precision? Why not to use `time.Now().UnixNano()` as is?

darioielardi: I always use milliseconds, but yours is a legit question, let me know if you want me to change that.

vmihailenco: Yes, please change it to use nanoseconds just like Go - that is what most Go users would expect (including me). Otherwise we just losing precision without saving any space...

darioielardi: Done, waiting for review.

orm/query.go
```diff
@@ -295,7 +295,7 @@ func (q *Query) Column(columns ...string) *Query {
 			continue
 		}
 
-		//TODO: remove
+		// TODO: remove
```
vmihailenco: just FYI I believe this was correct - space is not used to distinguish from normal comments

orm/model_table_struct.go
```diff
@@ -362,15 +363,29 @@ func (m *structTableModel) join(
...
+		switch {
+		case field.Type.Kind() == reflect.Int64:
+			value.Set(reflect.ValueOf(&unixNow))
+		case field.Type == timeType:
+			value.Set(reflect.ValueOf(&now))
+		}
+	} else {
+		switch {
+		case kind == reflect.Int64:
+			value.Set(reflect.ValueOf(unixNow))
```
vmihailenco: `value.SetInt` should be faster

vmihailenco:
Thanks for the PR - I am merging it and will do some changes myself. :+1:

PR #1548, opened 2020-03-20
Add possibility to override json library for jsonb fields

pymq:
After 9.1.1 release we encountered a few problems with new json library, for example `segmentio/encoding` doesn't support `omitempty` tags.

This PR introduce new JsonProvider API allowing user to overwrite library used for jsonb fields keeping `segmentio/encoding` library as default.

### DIFF
--- a/conv_test.go
+++ b/conv_test.go
@@ -10,9 +10,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/segmentio/encoding/json"
-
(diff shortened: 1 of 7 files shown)


vmihailenco:
@pymq thanks for the PR - I've changed it a bit in https://github.com/go-pg/pg/pull/1550. Please let me know if anything is wrong.

pymq:
Seems OK for me - thanks!

PR #1575, opened 2020-04-11
Call QueryHook-s for CopyFrom/CopyTo queries

### LINKED ISSUE #1564: Execute QueryHooks on CopyFrom/To
These hooks are very helpful for debugging query generation, latency measurement, etc. 
But currently it's not executed for CopyFrom/To.
If it's ok I can make PR to enable query hooks for copy commands.

xakep666:
Note for reviewer: Model field of QueryEvent filled if latest query argument is orm.TableModel. I found such approach in project. Is it good or it's better to not fill QueryEvent.Model?

Will close #1564


vmihailenco:
Thank you!

>Is it good or it's better to not fill QueryEvent.Model?

It looks like a nice idea - hopefully it will be useful.

PR #1580, opened 2020-04-15
patch update UpdateNotZero while use_zero flag

ramezmagdy:
patch UpdateNotZero function neglect use_zero flag in fields.

### DIFF
--- a/orm/update.go
+++ b/orm/update.go
@@ -167,7 +167,7 @@ func (q *updateQuery) appendSetStruct(fmter QueryFormatter, b []byte, strct refl
 
 	pos := len(b)
 	for _, f := range fields {
-		if q.omitZero && f.HasZeroValue(strct) {
+		if !f.hasFlag(UseZeroFlag) && q.omitZero && f.HasZeroValue(strct) {
 			continue
 		}
 


vmihailenco:
This would be a pretty substantial change that limits `UpdateNotZero` usability for fields with `use_zero` flag. It can be that current behavior is desired. There is following sentence in changelog

>>>UpdateNotNull is renamed to UpdateNotZero. As previously it omits zero Go values, but it does not take in account if field is nullable or not.

ramezmagdy:
i think it will not limit `UpdateNotZero` it will behave the same in case `use_zero` flag is not set for example if you have a model: 
```
type AdditionalContact struct {
	tableName       struct{}  `pg:"additional_contacts"`
	Id                     string    `pg:"id,pk,type:uuid,default:gen_random_uuid()"`
	AccountId        string    `pg:"account_id,type:uuid"`
	FirstName        string    `pg:"first_name"`
	LastName        string    `pg:"last_name"`
	IsEnabled        bool      `pg:"is_enabled,type:bool,use_zero"`
	SendAccountEmail bool      `pg:"send_account_email"`
	CreatedAt        time.Time `pg:"created_at"`
	UpdatedAt        time.Time `pg:"updated_at"`
}
```
and you want to set the IsEnabled to false only without updating the rest of the values with this change `UpdateNotZero` with produce the following query: 
update object : 
```
{
    "FirstName": "Takishi",
    "LastName": "Hamda",
    "Email": "t.hamada@main.com",
    "isEnabled": false,
    "SendAccountEmail": false
}
```
```
UPDATE additional_contacts AS "additional_contact" SET "first_name" = 'Takishi', "last_name" = 'Hamada', "email" = 't.hamada@main.com', "is_enabled" = FALSE, "updated_at" = '2020-04-16 08:51:01.544290498+00:00:00' WHERE "additional_contact"."id" = '132c0152-8e62-4f7a-9aa2-7cb5bcc48093'
```

frederikhors:
Same problem here, @vmihailenco.

I'm using `UpdateNotZero()` but sometimes I need to only "reset" a single column value and it would be useful to do it like this!

Please, can you reconsider?

@ramezmagdy Have you found another way to do it?

PR #1601, opened 2020-05-03
add embedded tag

letsfire:
use embedded tag same as Anonymous

```go
type Info struct {
    Name string
}

type Album struct {
    Info                         // Anonymous
    Info Info `pg:",embedded"`   // embedded，same as Anonymous
}
```


vmihailenco:
Can I ask why not to always use `Anonymous` / Go embedding?

letsfire:
Sometimes I don't want it to be anonymous

```
type Info struct {
    Name string
}

type Album struct {
    Info Info `pg:",embedded"`   // embedded，same as Anonymous
}

After json.Marshal，i want to be

{"info":{"name":""}}
```

PR #1613, opened 2020-05-19
Use SetSoftDeleteField in sliceTableModel

NathanBaulch:
This fixes two pointer related issues related to batch (slice) soft deleting.

1. Soft deleting a slice of pointer models with non-pointer soft delete fields.
2. Soft deleting a slice of non-pointer models with pointer soft delete fields.

Based on what I'm seeing in `structTableModel`, it would seem that using the recently introduced `Table.SetSoftDeleteField` function is the right way to go.

### DIFF
--- a/orm/model_table_slice.go
+++ b/orm/model_table_slice.go
@@ -3,10 +3,8 @@ package orm
 import (
 	"context"
 	"reflect"
-	"time"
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks!

PR #1624, opened 2020-05-29
check if valuer is nil before appending to query

nknordeen:
Ran into some unexpected panics when passing a nil pointer of type `*decimal.Decimal` to `func (db *baseDB) Exec(query interface{}, params ...interface{}) (res Result, err error) {`

Are there any instructions on contributing to this repo?  Sorry if I missed them


vmihailenco:
The method you've changed is a fast path that avoids using `reflect.Value`.But since now we have to use reflect.Value it makes sense to remove this "fast" version and use general processing that already checks for nils. I've sent https://github.com/go-pg/pg/pull/1626 that does that.

Thanks for the PR anyway!

PR #1636, opened 2020-06-17
Add Listener.Unlisten

jayschwa:
This enables a long-running `Listener` to unsubscribe from notification channels via the [`UNLISTEN` SQL command][1].

Reads and writes to the `Listener.channels` slice are now protected by a mutex to prevent data races.

This patch is for go-pg v8 because that is what our downstream project is using.

[1]: https://www.postgresql.org/docs/current/sql-unlisten.html

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -42,6 +42,9 @@ type Listener struct {
 }
 
 func (ln *Listener) String() string {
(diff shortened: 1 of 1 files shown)


vmihailenco:
Looks good -  thanks!

PR #1652, opened 2020-07-07
RegisterAppender and RegisterScanner only panic if the type was already explicitly registered

marcel:
When registering a scanner or an appender, only panic if the type has 
already been explicitly registered with `types.RegisterScanner` or
`types.RegisterAppender`, rather than if it was implicitly set to a default as a
side-effect of `types.Appender` or `types.Scanner` being called.

This avoids a race where a previous call to `types.Appender` or
`types.Scanner` which were invoked as a side-effect assigned a default
`Appender/Scanner` which then causes a subsequent call to
`Register{Appender,Scanner}` to panic. This can happen if, for example,
`orm.GetTable` is invoked before an `init()` function that calls Register
gets invoked due to unlucky package import ordering, since `orm.GetTable`
tries to look up the type's `Appender/Scanner`.

Example:
```go
  // This package gets imported first
  package bar

  type MyStruct struct {
    MyType MyType
  }

  var Table = orm.TableFor(&MyStruct{})
```
```go
  // This package gets imported second
  package foo

  type MyType struct {}

  func init() {
    // This panics because orm.TableFor called types.Appender and,
    finding none, set a default
    types.RegisterAppender(&MyType{}, ...)
  }
```

### DIFF
--- a/types/append_value.go
+++ b/types/append_value.go
@@ -55,6 +55,7 @@ func init() {
 }
 
 var appendersMap sync.Map
+var registeredAppendersMap sync.Map
 
 // RegisterAppender registers an appender func for the value type.
 // Expecting to be used only during initialization, it panics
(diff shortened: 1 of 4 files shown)


vmihailenco:
When called `var Table = orm.TableFor(&MyStruct{})` caches the appender. So `types.RegisterAppender` will overwrite the appender, but `Table` will continue to use old appender. That is potentially very confusing - I think current behavior is preferable.

CyborgMaster:
@marcel, I just ran into this exact issue this morning.  Did you find a workaround?

marcel:
> @marcel, I just ran into this exact issue this morning.  Did you find a workaround?

I've since changed to a different database so did not need to address the problem. A janky work around would be to use an underscore import in every file that is loaded lexicographically prior to force the init that calls Register to be executed first.

CyborgMaster:
Yeah, I suppose that is what I'll do for now.

PR #1698, opened 2020-08-25
 using go.opentelemetry.io/otel @v0.11.0

burner-account:
see https://github.com/open-telemetry/opentelemetry-go/releases/tag/v0.11.0

Only relevant change was that they moved otel/api/kv to the otel/label package and that they now no longer use grpc status codes but their own clone of it.

Most tests run OK, access via unix socket and redis-based test (notifications) do not pass. Those failing test should be a result of my half-assed set up test environment.

PR reason: broke my toys running go get -u ./... and so i had to fix it


burner-account:
This PR contains minor version updates to other packages as `go.opentelemetry.io/otel`, too.

vmihailenco:
Thanks!

PR #1728, opened 2020-09-14
Dont Swallow Before Query Errors

CyborgMaster:
If the before query hook returned an error, we never returned it to the outside service.  The After Query hook was hiding it.


vmihailenco:
Thanks

CyborgMaster:
Thanks for the quick merge!

PR #1745, opened 2020-09-25
dependency update and workaround for opentelemetry api change

burner-account:
The pull request of the renovate bot is broken. #1744 

This PR should address the immediate problem arising from keeping opentelemetry up-to-date.

Anyway, without further knowledge i would guess that there is  something wrong about the approach of using opentelemetry. At least the devs there thought nobody would run into the problem addressed by this PR. [their issue](https://github.com/open-telemetry/opentelemetry-go/issues/300)

Was having opentelemetry in go-pg 'a nice idea' or a 'must have'? Maybe it could be removed from go-pg itself.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,10 +13,10 @@ require (
 	github.com/vmihailenco/bufpool v0.1.11
 	github.com/vmihailenco/msgpack/v5 v5.0.0-beta.1
 	github.com/vmihailenco/tagparser v0.1.1
-	go.opentelemetry.io/otel v0.11.0
+	go.opentelemetry.io/otel v0.12.0
 	golang.org/x/crypto v0.0.0-20200820211705-5c72a883971a // indirect
-	golang.org/x/net v0.0.0-20200904194848-62affa334b73 // indirect
-	golang.org/x/sys v0.0.0-20200918174421-af09f7315aff // indirect
+	golang.org/x/net v0.0.0-20200925080053-05aa5d4ee321 // indirect
+	golang.org/x/sys v0.0.0-20200923182605-d9f96fdee20d // indirect
 	google.golang.org/appengine v1.6.6 // indirect
 	google.golang.org/protobuf v1.25.0 // indirect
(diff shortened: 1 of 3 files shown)


vmihailenco:
Thanks for fixing this.

In general this should not be a problem as long as you don't run `go get -u`. Even then you can stay on old version if something goes wrong. That is why we have go.mod with ability to lock the version :)

burner-account:
You say i should accept best practices and use the tools i am given instead of hoping the world will bow to my lack of motivation to write more fine-grained build scripts? How dare you :)

Yeah, i will do that. But i really meant it. If there is an easy way to decouple go-pg from opentelemetry, it is worth investigating from my pov.

vmihailenco:
OpenTelemetry hits v1 in November or so. And no, there is no way to achieve the same outside of go-pg - go-pg records some internal details like time to establish connection, retries etc.

burner-account:
Their API hitting v1 in the next months or so is one thing, i get that train of thought.

But there should be a way (with little performance overhead, more of structural overhead) exposing hooks in go-pg which could be used in - let's say - go-pg/opentelemetry to provide the opentelemetry stuff.

This way one day there could be go-pg/flavor-non-opentelemetry or something alike.

Anyway, i am not too attached to the decoupling idea, so it is all fine with me. It is simply me stumbling over the (at least perceived) fact that opentelemetry is not married to the purpose of go-pg, but it is married to its code.

Thank you for your quick merge and for your good work in general.

edit: go-pg/flavor-non-opentelemetry means "using another telemetry system"

vmihailenco:
>go-pg/flavor-non-opentelemetry means "using another telemetry system"

OpenTelemetry is open and backend agnostic - you can write your own backend. Which is exactly what I did - https://uptrace.dev :) AFAIK there are no alternatives (edit: for Otel, not Uptrace).

Regarding hooks I've added the answer in [docs](https://docs.uptrace.dev/tracing/#opentelemetry-overhead). The short answer - OpenTelemetry is a specialized hook system for collecting telemetry. The data can be exported wherever you want.

PR #1803, opened 2020-12-18
Add go-pg-monitor to README ecosystem section

hypnoglow:
Hi! I made a package to export db client stats as Prometheus metrics. We already use it for some of our services.
If you see a fit, we can add this to README ecosystem section, maybe someone will find it useful.
Thanks.


vmihailenco:
Thanks. Later we plan to add more metrics using OpenTelemetry but this looks good already.

PR #1814, opened 2021-01-28
Replace context for Rollback when it is canceled

### LINKED ISSUE #1813: Current transaction is aborted, commands ignored until end of transaction block
## Expected Behavior

Error in the transaction but client remains stable

## Current Behavior

When the context of a transaction gets cancelled, the following errors happen:

### Client-side error:
 
These errors occur:

```
tx.Rollback failed: ERROR #57014 canceling statement due to user request
```

```
ERROR #57014 canceling statement due to user request
```

Then during at least one minute, all transactions fail with:

```
ERROR #25P02 current transaction is aborted, commands ignored until end of transaction block
```

image

### Server-side error

In a loop:

```
user=user,db=db,app=[unknown],client=xx.xxx.xx.xx LOG:  08006: could not receive data from client: Connection reset by peer
```

## Possible Solution

Perhaps when a transaction context gets cancelled (e.g, timeout), rollback of a transaction should be run anyway?

If you have any clue about what can be done in go-pg, we'd be pleased to help implement it.

## Steps to Reproduce

This error is quite random. At first it used to happen because too many connections were opened on the server. Now this problem is fixed (I am far under the client connections limit) but now it's fixed, but this still happens sometimes.

vmihailenco:
Fixes https://github.com/go-pg/pg/issues/1813

### DIFF
--- a/internal/context.go
+++ b/internal/context.go
@@ -1,0 +1,26 @@
+package internal
+
+import (
+	"context"
+	"time"
+)
+
+type UndoneContext struct {
+	context.Context
+}
+
+func UndoContext(ctx context.Context) UndoneContext {
+	return UndoneContext{Context: ctx}
+}
+
+func (UndoneContext) Deadline() (deadline time.Time, ok bool) {
+	return time.Time{}, false
+}
+
+func (UndoneContext) Done() <-chan struct{} {
+	return nil
+}
+
+func (UndoneContext) Err() error {
+	return nil
+}
--- a/tx.go
+++ b/tx.go
@@ -339,7 +339,7 @@ func (tx *Tx) Commit() error {
 
 // Commit commits the transaction.
 func (tx *Tx) CommitContext(ctx context.Context) error {
-	_, err := tx.ExecContext(ctx, "COMMIT")
+	_, err := tx.ExecContext(internal.UndoContext(ctx), "COMMIT")
 	tx.close()
 	return err
 }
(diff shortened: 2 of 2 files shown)


Barben360:
Hello,

There is some flaw in the method you are using: if the context cancels between the call to `healthyContext`, then the problem will still occur. Wouldn't it be even simpler always to use `context.Background()` whatever the error is?

vmihailenco:
It would "break" tracing which uses context for passing active span.

I guess we need some way to remove timeout from the context, e.g.

```go
type UndoneContext struct {
    context.Context
}

func (ctx UndoneContext) Done() <-chan struct{} {
    return nil
}

// Same for Deadline and Err
```

And apply it to the context passed to Rollback/Close. PRs are welcome :)

Barben360:
Yes indeed tracing is a problem. What I'm used to doing instead of "removing timeouts", which may not be the only cause of cancellation, is taking a fresh new context and extracting and re-injecting from incoming context only what's necessary for tracing or other debug features. Do you think that it'd be possible in go-pg's infrastructure? All you need is to have access to your tracing library where you want do to that.

I'm actually already doing this in some cases not to loose trace. In the following example, I need to create a new context and I don't want to loose the trace:

```go
if Trace {
	ctx = trace.ContextWithRemoteSpanContext(context.Background(), trace.SpanContextFromContext(ctx))
} else {
	ctx = context.Background()
}
```

This is used with `go.opentelemetry.io/otel/trace`, which luckily is the tracer you are using in go-pg!

vmihailenco:
@Barben360 I've updated this PR to use `UndoneContext`. Let me know if you spot anything wrong.

Barben360:
@vmihailenco Nice trick, that seems ok to me!

vmihailenco:
:+1: Merging and releasing then.

PR #1815, opened 2021-01-29
Use safe context for transaction rollback

### LINKED ISSUE #1813: Current transaction is aborted, commands ignored until end of transaction block

Barben360:
Closes #1813 

Here is a draft for a fix proposition.

The 10 seconds for timeout are totally arbitrary and so maybe you'd prefer to have something configurable.

The trace span injection should not be problematic even if no span is in the context.

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -6,10 +6,12 @@ import (
 	"io"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/go-pg/pg/v10/internal"
 	"github.com/go-pg/pg/v10/internal/pool"
 	"github.com/go-pg/pg/v10/orm"
+	"go.opentelemetry.io/otel/trace"
 )
 
 // ErrTxDone is returned by any operation that is performed on a transaction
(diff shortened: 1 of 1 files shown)


tx.go
```diff
@@ -85,16 +87,22 @@ func (tx *Tx) Begin() (*Tx, error) {
 func (tx *Tx) RunInTransaction(ctx context.Context, fn func(*Tx) error) error {
 	defer func() {
 		if err := recover(); err != nil {
-			if err := tx.RollbackContext(ctx); err != nil {
-				internal.Logger.Printf(ctx, "tx.Rollback panicked: %s", err)
+			safeCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			safeCtx = trace.ContextWithRemoteSpanContext(safeCtx, trace.SpanContextFromContext(ctx))
```
vmihailenco, line 92: This should be `trace.ContextWithSpan` since remote span context may have some implications (like with sampling).

But I think I prefer `UndoneContext` since it preserves all values in context...

PR #1821, opened 2021-02-05
Fix logger prefix

AnatolyRugalev:
Well, one thing I know for sure is that there is no reason we should have `redis: ` prefix in go-pg logger


vmihailenco:
Thanks :)

PR #1831, opened 2021-02-10
(#1830) Fixed notice during startup error.

### LINKED ISSUE #1830: pg: unknown startup message response: 'N'
Issue tracker is used for reporting bugs and discussing new features. Please use
[Discord](https://discord.gg/rWtp5Aj) or [stackoverflow](https://stackoverflow.com) for supporting
issues.

## Expected Behavior

go-pg would do nothing or would not return an error.

## Current Behavior

go-pg returns an error `pg: unknown startup message response: 'N'`

## Possible Solution

Handle `N` messages during startup when the `Code:` is `42622`.

## Steps to Reproduce

```
	options := &pg.Options{
		Network:         "tcp",
		Addr:            os.Getenv("POSTGRES_HOST") + ":5432",
		User:            os.Getenv("POSTGRES_USER"),
		Password:        os.Getenv("POSTGRES_PASSWORD"),
		Database:        os.Getenv("POSTGRES_DB"),
		ApplicationName: "i am just a really really super long application name so that i make a notice during startup",
	}
	db := pg.Connect(options)

	require.NoError(t, db.Ping(context.Background()), "must ping database") // Error occurs here.
```

## Context (Environment)

When running tests in CI against a database I will append `t.Name()` to my application name to see what each test is doing. If a test name is too long this error occurs when connecting to the database.

image

## Possible Implementation

I'm happy to make a PR for this and resolve it but wanted to open an issue for visibility until then. 
- Could simply truncate the application name in go-pg.
- Add handling of notices during startup for that particular code.

elliotcourant:
This adds handling of Notice messages during startup and will log the messages for the client to see so they may take action.

Resolves #1830

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 
 	"mellium.im/sasl"
 
@@ -116,6 +117,15 @@ func (db *baseDB) startup(
 			case readyForQueryMsg:
 				_, err := rd.ReadN(msgLen)
 				return err
+			case noticeResponseMsg:
+				// If we encounter a notice message from the server then we want to try to log it as it might be
+				// important for the client. If something goes wrong with this we want to fail. At the time of writing
+				// this the client will fail just encountering a notice during startup. So failing if a bad notice is
+				// sent is probably better than not failing, especially if we can try to log at least some data from the
+				// notice.
+				if err := db.logStartupNotice(rd); err != nil {
+					return err
+				}
 			case errorResponseMsg:
 				e, err := readError(rd)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


messages.go
```diff
@@ -178,6 +188,51 @@ func (db *baseDB) auth(
...
+
+		// If they key byte (the type of field this data is) is 0 then that means we have reached the end of the notice.
+		// We can break our loop here and throw our message data into the logger.
+		if fieldType == 0 {
+			break
+		}
+
+		// Read until the next null byte to get the data for this field. This does include the null byte at the end of
+		// fieldValue so we will trim it off down below.
+		fieldValue, err := rd.ReadSlice(0)
```
vmihailenco: Can't you use `readString(rd)` here?

messages.go
```diff
@@ -178,6 +188,51 @@ func (db *baseDB) auth(
...
+		// Just throw the field type as a string and its value into an array.
+		// Field types can be seen here: https://www.postgresql.org/docs/13/protocol-error-fields.html
+		// TODO This is a rare occurrence as is, would it be worth adding something to indicate what the field names
+		//  are? Or is PostgreSQL documentation enough for a user at this point?
+		message = append(message, fmt.Sprintf("%s: %s", string(fieldType), string(fieldValue[:len(fieldValue)-1])))
+	}
+
+	// Tell the client what PostgreSQL told us. Warning because its probably something the client should at the very
+	// least adjust.
+	internal.Warn.Printf("notice during startup; %s", strings.Join(message, ", "))
```
vmihailenco: Not sure if `;` is intentional or you wanted `:`

messages.go
elliotcourant: I wasn't aware of that method, but yeah if it will read until the next null character then that's exactly what i need.

messages.go
elliotcourant: I was indecisive which to use? I can use `:` if its preferred

vmihailenco: Whatever you prefer :)

messages.go
vmihailenco: :+1: this should be ready to go once this is fixed

elliotcourant:
Tweaked for feedback and squashed the commit!

vmihailenco:
Thanks and sorry for bugging you about it :)

elliotcourant:
You didn't bug me at all! Thank you!

PR #1843, opened 2021-03-09
Fix otel db.system name

mattnibs:
According to the open [telemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes):

> db.system MUST be one of the following or, if none of the listed values apply, a custom value:

Down the list the name for postgres reads as "postgresql".

Add the remaining "ql" characters and conform the expectations of the spec.


vmihailenco:
Thanks :+1: This is a disrupting change for tracing systems but I am not sure what is the best way to handle it... guess it is better to do it now than later.

vmihailenco:
Tagged pgotel v0.3.1

Issue #1845, opened 2021-03-12
Order() doesn't support `COLLATE <name> ASC|DESC` syntax.


jeffreydwalter:
Issue tracker is used for reporting bugs and discussing new features. Please use
[Discord](https://discord.gg/rWtp5Aj) or [stackoverflow](https://stackoverflow.com) for supporting
issues.

PostgreSQL supports icu collation, but the go-pg library's Order() method does not have support for it and so does the wrong thing.

## Expected Behavior
```
If I have the following: 
q.Order(`product.name COLLATE icu_natural DESC`)

it should produce: 
ORDER BY "product"."name" COLLATE icu_natural DESC
```

## Current Behavior
```
If I have the following: 
q.Order(`product.name COLLATE icu_natural DESC`)

it produces: 
ORDER BY "product"."name COLLATE icu_natural DESC"
```

## Possible Implementation

PR: https://github.com/go-pg/pg/pull/1846

vmihailenco:
`Order` is meant to accept data from end users (e.g. via URL query) and automatically quotes it against SQL injections. In all other cases `OrderExpr` should be used.

Could you confirm that you really want to accept variable `COLLATE` from users? It is a bit unusual...

dsonck92:
@vmihailenco Hmm, I can see that viewpoint now. Perhaps update the docs on the Order function to highlight this usage as it's currently not that obvious. I mean, it's an Order function that quotes the column name, that much is clear. But perhaps adding "Which means userdata can be passed directly which is useful for handling sorting payloads from requests" or something in that line.

On the other hand, it may be useful for frontends to specify the collation as users may expect a different sorting based on their locale as deduced by the frontend whereby this feature results in less code between the frontend and go-pg layer as it could send a request including the COLLATE bit.

When I look at `NULLS FIRST/LAST`, while this is perfectly valid sent by the frontend, I expect this API not being directly exposed to users by the frontend but wrapped around some "unassigned first/last" checkbox. This could be the same for `COLLATE` where the frontend may send this, but also not directly expose it similar to NULLS. However, if this is properly handled and quoted, then the effort in the frontend would be very minimal:

```typescript
const sorting = [`${column} COLLATE ${user.locale} ${direction}`]
```
where `user.locale` would be some local state specifying whichever locale information is normal for the user. That's just my 2cts

jeffreydwalter:
[...]

PR #1846, opened 2021-03-12
Issue 1845: Add support for COLLATE in ORDER BY clause.

jeffreydwalter:
https://github.com/go-pg/pg/issues/1845


orm/query.go
```diff
@@ -709,9 +710,8 @@ loop:
 		if ind != -1 {
 			field := order[:ind]
 			sort := order[ind+1:]
-			switch internal.UpperString(sort) {
-			case "ASC", "DESC", "ASC NULLS FIRST", "DESC NULLS FIRST",
-				"ASC NULLS LAST", "DESC NULLS LAST":
+			reg := regexp.MustCompile(`(?i)^((ASC|DESC)( NULLS FIRST| NULLS LAST)?)|(COLLATE [a-zA-Z1-9_]+ (ASC|DESC))$`)
```
dsonck92: Now, I don't work with `COLLATE` much but I assume that a `COLLATE` can still specify `NULLS FIRST/LAST` whereas this regex will only allow `NULLS FIRST/LAST` or `COLLATE` at a time.

Also, in the Postgres docs, I see that collations are quoted as identifiers, so probably it's proper to extract the `COLLATE` name as well and specify it as `types.Ident`.

I might play with Postgres later to verify my first question.

jeffreydwalter: Same here @dsonck92, I updated my PR to support the case you mentioned.

dsonck92: Meanwhile I was so close to a fix already (though it bypasses the regex so I may share it for sake of posterity to let @vmihailenco decide which is preferred). But yes, the code is now covering more features which I think is best. AFAICS, LGTM

dsonck92:
For sake of posterity and choice (and not to let my few minutes work go in vain)

```golang
func extractCollate(sort string) (sortOut, collate string, err error) {
	ind := strings.Index(sort, " ")
	if ind != -1 && internal.UpperString(sort[:ind]) == "COLLATE" {
		// chomp "COLLATE "
		sortOut = sort[ind+1:]
		ind = strings.Index(sortOut, " ")
		if ind != -1 {
			collate = sortOut[:ind]
			sortOut = sortOut[ind+1:]
			return sortOut, collate, nil
		} else {
			// Failure to get the collation
			return "", "", errors.New("invalid collate")
		}
	}
	return sort, "", nil
}

// Order adds sort order to the Query quoting column name. Does not expand params like ?TableAlias etc.
// OrderExpr can be used to bypass quoting restriction or for params expansion.
func (q *Query) Order(orders ...string) *Query {
loop:
	for _, order := range orders {
		if order == "" {
			continue
		}
		fmt.Println("Order:", order)
		ind := strings.Index(order, " ")
		if ind != -1 {
			field := order[:ind]
			sort, collate, err := extractCollate(order[ind+1:])
			if err != nil {
				continue
			}

			switch internal.UpperString(sort) {
			case "ASC", "DESC", "ASC NULLS FIRST", "DESC NULLS FIRST",
			"ASC NULLS LAST", "DESC NULLS LAST":
				if collate != "" {
					q = q.OrderExpr("? COLLATE ? ?", types.Ident(field), types.Ident(collate), types.Safe(sort))
				} else {
					q = q.OrderExpr("? ?", types.Ident(field), types.Safe(sort))
				}
				continue loop
			}
		}

		q.order = append(q.order, fieldAppender{order})
	}
	return q
}
```
Which should be largely compatible with the latest test case except that `COLLATE` is always uppercase

---

Noting that I cannot judge the readability of this vs regex. I did extract the logic out of the function, though technically it can also be embedded. The error can also simply be a `return sort, ""` as the switch will catch illegal constructions anyways.

jeffreydwalter:
@dsonck92 in general I like your solution better. It's more straight forward and probably more performant without the regex and looping. 👍 The only issue I see is that it fails the unit tests because it doesn't respect the casing of `COLLATE`, which the unit test is written to do. Obviously an easy fix.
image

jeffreydwalter:
@dsonck92 I fixed the issue with COLLATE being hardcoded and thus failing the unit tests and pushed your code for this PR. I did some quick and dirty performance testing and your solution was an order of magnitude faster. 👍 Thanks for that.

dsonck92:
I figured, considering that regex is powerful but not necessarily that fast. And considering the old code used some "internal" functions that probably cater for speed, yes. Well, I'm glad it was useful so I didn't write code for nothing :sweat_smile:. And yeah, I was thinking about the collate casing bit, but at the time I didn't extract it to a function yet so it looked a bit weird.

But I must say seeing it there in the code, it's not too bad to have the extra variable in there. Now lets hope it can be merged in.

vmihailenco:
Closing based on the discussion in https://github.com/go-pg/pg/issues/1845. The code in this PR is good, but no code is even better & faster :)

vmihailenco:
PS We can reconsider if there is real interest in using it. Thanks for the PR anyway :+1:

PR #1885, opened 2021-05-21
Mark more errors as fatal

### LINKED ISSUE #1879: Canceling context randomly breaks everything
There seem to be an important problem of concurrency in go-pg.

When there are concurrent calls and a context gets cancelled in one of them, this seems to break the whole connection pool.

To reproduce it, I have written the following test:

```go
func TestConcurrencyAndUserCancel(t *testing.T) {
	const concurrency = 10
	const iterations = 1000
	const cancelProbability = 0.1

	app, stop := InitScenario(t)
	defer stop()
	
	ctxBkg := context.Background()

	cancelWaitMin := 1 * time.Microsecond
	cancelWaitMax := 5 * time.Millisecond

	cancelProbabilityScaled := int(cancelProbability * 1000000)

	obj := models.SomeObj{}

	wg := sync.WaitGroup{}
	wg.Add(concurrency)
	for i := 0; i < concurrency; i++ {
		go func(wg *sync.WaitGroup, i int) {
			defer wg.Done()
			for j := 0; j < iterations; j++ {
				// Are we gonna cancel our call ?
				cancelWait := time.Duration(0)
				val := rand.Intn(1000000)
				if val <= cancelProbabilityScaled {
					cancelWait = time.Duration(rand.Int63n(int64(cancelWaitMax-cancelWaitMin))) + cancelWaitMin
					t.Logf("Canceling add statement in %v", cancelWait)
				}

				// Adding object
				ctx := ctxBkg
				cancel := func() {}
				if cancelWait != 0 {
					ctx, cancel = context.WithTimeout(ctx, cancelWait)
				}
				objCopy := obj
				err := app.Obj.Add(ctx, &objCopy)
				cancel()
				if err != nil {
					if cancelWait == 0 {
						t.Errorf("An error occurred for concurrent %d - iteration %d - call 1: %v", i, j, err)
					}
					continue
				}

				// Editing object
				objCopy.SomeField = "some string"
				err = app.Obj.Edit(ctxBkg, &objCopy)
				if err != nil {
					t.Errorf("An error occurred for concurrent %d - iteration %d - call 3: %v", i, j, err)
				}

				// deleting object
				err = app.Obj.Delete(ctxBkg, objCopy.Id)
				if err != nil {
					t.Error(err)
				}
			}
		}(&wg, i)
	}
	doneChan := make(chan bool)
	go func(doneChan chan bool) {
		wg.Wait()
		close(doneChan)
	}(doneChan)

	select {
	case <-doneChan:
	case <-time.After(10 * time.Minute):
		t.Error("Test timeout")
	}
}
```

In this test, 10 concurrents are making each 1000 sequential INSERT/EDIT/DELETE (which are all run inside `RunInTransaction()`). With a probability of 10%, the context of the transaction of the INSERT is cancelled with a random timeout.

I'm using a pool of 10 connections.

- With a probability of context cancel of 0%, this test always passes, whatever the concurrency is between 1 and 10.
- With a probability of context cancel of 10%, this test breaks some times with a concurrency of 5.
- With a probability of context cancel of 10%, this test breaks nearly all the time with a concurrency of 10.

What happens is that after a context cancel (one of the many, randomly), all subsequent calls of all concurrents break with:

```
tx.go:97: tx.Rollback failed: pg: Conn is in a bad state: write tcp 172.24.0.4:37304->172.24.0.3:5432: i/o timeout
```

and the client and server never communicate again.

This is extremely annoying because the whole application may become indefinitely unresponsive until restarted, especially when there is a high charge on it.

vmihailenco:
Fixes https://github.com/go-pg/pg/issues/1879

### DIFF
--- a/error.go
+++ b/error.go
@@ -43,7 +43,16 @@ func isBadConn(err error, allowTimeout bool) bool {
 		return false
 	}
 	if pgErr, ok := err.(Error); ok {
-		return pgErr.Field('S') == "FATAL"
+		switch pgErr.Field('V') {
+		case "FATAL", "PANIC":
+			return true
+		}
+		switch pgErr.Field('C') {
+		case "25P02", // current transaction is aborted
+			"57014": // canceling statement due to user request
+			return true
+		}
+		return false
 	}
 	if allowTimeout {
 		if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
--- a/tx_test.go
+++ b/tx_test.go
@@ -135,7 +135,7 @@ var _ = Describe("Tx", func() {
 		Expect(count).To(Equal(1))
 
 		err = tx1.Commit()
-		Expect(err).NotTo(HaveOccurred()) // actually ROLLBACK happens here
+		Expect(err).To(HaveOccurred()) // actually ROLLBACK happens here
 
 		_, err = tx2.QueryOne(pg.Scan(&count), "SELECT COUNT(*) FROM test_copy_from")
 		Expect(err).NotTo(HaveOccurred())


vmihailenco:
This probably should be just:

```go
	if pgErr, ok := err.(Error); ok {
		return !err.IntegrityViolation()
	}
```

But it is a more aggressive fix that could hurt performance if we don't whitelist enough errors...

PR #1899, opened 2021-07-22
int[] type to array conversion

inv2004:
Pg 13.0 uses 1007 internal type to represent int[] type

Added the type into column.go to support the conversion
It still uses `scanInt64Array` like int8[]

Regards,


vmihailenco:
Thanks

PR #1906, opened 2021-08-07
Add possibility to specify custom `join_fk` for "has-one" relations.

### LINKED ISSUE #1812: I cant use "join_fk" with "has-one" Relation
I want to use "has-one" relation. Example;

```
type UserDB struct {
	ID int `json:"-" 
	UserID      string    `json:"userID" pg:",unique"`
	 // Other user fields
}

type UserDetailDB struct {
	ID int `json:"-" `
	UserID      string    `json:"userID" pg:",unique"`
	UserDB      *UserDB `json:"user,omitempty" pg:"rel:has-one,fk:user_id,join_fk:user_id"`
      // Other User Detail fields
}

var userDetailDB []UserDetailDB
db.Model(&userDetailDB).Column("user_detail_db.*").Relation("UserDB") ....
```

When I do this, it creates a query like 

LEFT JOIN" user_dbs "AS" user_db "ON" user_db "." Id "=" user_detail_db "." User_id "

While "user_detail_db" gets "user_id" as I want, "user_db" gets "id". I want "user_id" in both tables. 

What can I do ?

xakep666:
Will close #1812.

### DIFF
--- a/orm/join_test.go
+++ b/orm/join_test.go
@@ -32,12 +32,36 @@ type BelongsTo struct {
 	JoinTestId int
 }
 
+type HasOneNonPK struct {
+	tableName struct{} `pg:"HasOneNonPK,alias:HasOneNonPK"`
+
+	Id int
+
+	CustomKey string `pg:"custom_key"`
+}
+
+type NonPKJoinTest struct {
+	tableName struct{} `pg:"NonPKJoinTest,alias:NonPKJoinTest"`
+
+	Id int
+
+	CustomHasOneKey string  `pg:"custom_has_one_key"`
+	CustomHasOne    *HasOneNonPK `pg:"rel:has-one,fk:custom_has_one_key,join_fk:custom_key"`
+}
+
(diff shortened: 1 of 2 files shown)


orm/join_test.go
```diff
@@ -23,6 +26,8 @@ type HasOne struct {
 
 	HasOne   *HasOne
 	HasOneId int
+
+	CustomKey string `pg:"custom_key"`
```
vmihailenco: Please don't change existing tests - otherwise it is not clear whether the old code is working the same or not. This should be a new separate test or an example.

xakep666: @vmihailenco is it ok to just duplicate old test and modify it?

vmihailenco: Yes, that is perfectly fine :+1:

vmihailenco:
Thanks - this looks good :+1:  I will need to find some time to migrate the build system to Github actions and then will merge this.

PR #1917, opened 2021-08-24
Replace Travis with Github actions

vmihailenco:


.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version: [1.16.x, 1.17.x]
+
+    services:
+      postgres:
+        image: postgres
```
elliotcourant: Would you want to make the postgres version a matrix as well?

Verify behaviors are valid for 11, 12 and 13? Or are there not significant enough differences right now?

vmihailenco: There is no code that depends on Postgres version so it is probably not worth the trouble at the moment...

PR #1918, opened 2021-08-25
Add support for ScramSha256Plus

### LINKED ISSUE #1483: pg fails to connect to postgres with SASL error
Hi,

I'm running postgres 11.5 (via RDS, though I don't think it matters) and when I try to connect with pg I get the following error:
```
"pg: SASL: got \"SCRAM-SHA-256-PLUS\", wanted \"SCRAM-SHA-256\""
```

Is this possibly a known issue?

Thanks,

---

**edit:** putting in my notes for anyone else who finds this later. SCRAM-SHA-256-PLUS is used when the postgres server requires TLS to connect to it (as I have configured my RDS instance to do). Since go-pg doesn't support this method yet the 2 current options are to:
- update postgres to allow plaintext connections / stop requiring TLS
- or, downgrade the role's password to MD5 instead of SCRAM

I have opted for the latter solution. This can itself be done in one of two ways:
- Update your postgres [authentication settings](https://www.postgresql.org/docs/12/runtime-config-connection.html#RUNTIME-CONFIG-CONNECTION-AUTHENTICATION) and set the inappropriately named `password_encryption` parameter to `md5` (if you're using AWS RDS this is controlled via a parameter group). Then update the role's password via psql:
    ```
    postgres=> alter role myrole with password 'password';
    ```
- or, directly update the role's password. To allow dumping/restoring backups, postgres lets you assign a raw md5/scram value to the role's password column. The md5 format postgres uses is the string "md5" followed by the md5 hash of the password concatenated with the role's name. So for a role "myrole" and password "password", you can run:
    ```
    $ echo -n "passwordmyrole" | md5sum | awk '{print "md5"$1}'
    md519488a15bff1b244feb52d621ac8b965
    ```
    Then set this as the password value via psql:
    ```
    postgres=> alter role myrole with password 'md519488a15bff1b244feb52d621ac8b965';
    ```

Hope this helps someone

vmihailenco:
Fixes https://github.com/go-pg/pg/issues/1483

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
+# Maintenance mode
+
+go-pg is in a maintenance mode and only critical issues are addressed. New development happens in
+[**Bun**](https://bun.uptrace.dev/guide/pg-migration.html) repo which offers similar functionality
+but works with PostgreSQL, MySQL, and SQLite.
+
+---
+
 # Changelog
-
-> :heart:
-> [**Uptrace.dev** - All-in-one tool to optimize performance and monitor errors & logs](https://uptrace.dev)
-
-**Important**. Please check [Bun](https://bun.uptrace.dev/guide/pg-migration.html) - the next
-iteration of go-pg built on top of `sql.DB`.
 
 ## v10.10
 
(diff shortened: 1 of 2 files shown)


Neustradamus:
@vmihailenco: Good job!

SCRAM-SHA-256 has been added in PostgreSQL 10
SCRAM-SHA-256-PLUS variant (with TLS Binding) has been added in PostgreSQL 13
SCRAM-SHA-256 is selected by default in PostgreSQL 14

Links:
- https://www.postgresql.org/about/news/postgresql-10-released-1786/
- https://www.postgresql.org/about/news/postgresql-13-released-2077/
- https://www.postgresql.org/about/news/postgresql-14-released-2318/
- https://techcommunity.microsoft.com/t5/azure-database-for-postgresql/how-to-securely-authenticate-with-scram-in-postgres-13/ba-p/1548319

PR #1930, opened 2021-11-20
fix: Fixed `NULL` not being formatted for slices/maps.

### LINKED ISSUE #1908: incorrect query marshalling the nil []string to null
```
create table settings
(
    aspects      jsonb       default '[]'::jsonb,
);
```

```golang
var a []string
Db.Query( &ss, `select * from settings where aspects @> coalesce(?, '[]'::jsonb)`, a)
```
the query msg will become: 
```
select * from settings where aspects @> coalesce('null', '[]'::jsonb)
```
not what I expect:
```
select * from settings where aspects @> coalesce(null, '[]'::jsonb)
```
only the literal `nil` work:
```golang
var a []string
Db.Query( &ss, `select * from settings where aspects @> coalesce(?, '[]'::jsonb)`, nil)
```
results:
```
select * from settings where aspects @> coalesce(null, '[]'::jsonb)
```

elliotcourant:
When a query parameter was a nil slice or a nil map, the value `NULL`
would not be appended to the query. Instead `'null'` would be appended
to the query which is not correct.

Resolves #1908

### DIFF
--- a/types/append_value.go
+++ b/types/append_value.go
@@ -132,8 +132,14 @@ func ptrAppenderFunc(typ reflect.Type) AppenderFunc {
 }
 
 func appendValue(b []byte, v reflect.Value, flags int) []byte {
-	if v.Kind() == reflect.Ptr && v.IsNil() {
-		return AppendNull(b, flags)
+	switch v.Kind() {
+	case reflect.Ptr, reflect.Slice, reflect.Map:
+		// When the value is a pointer, slice or map; then we can assert if the value provided is nil.
+		// When the value is properly nil; then we can append NULL instead of a string representation of the provided
+		// value.
+		if v.IsNil() {
+			return AppendNull(b, flags)
+		}
 	}
 	appender := Appender(v.Type())
 	return appender(b, v, flags)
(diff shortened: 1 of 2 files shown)


types/append_value.go
```diff
@@ -132,8 +132,14 @@ func ptrAppenderFunc(typ reflect.Type) AppenderFunc {
...
+	switch v.Kind() {
+	case reflect.Ptr, reflect.Slice, reflect.Map:
+		// When the value is a pointer, slice or map; then we can assert if the value provided is nil.
+		// When the value is properly nil; then we can append NULL instead of a string representation of the provided
+		// value.
+		if v.IsNil() {
+			return AppendNull(b, flags)
+		}
 	}
 	appender := Appender(v.Type())
```
vmihailenco, line 144: The idea is that you should be able to call `Appender` once for a type and then use it to append different values of the given type (nil or non-nil). So the function returned by appender should handle nil values, not `appendValue` itself.

I guess it is not worth fixing this properly so let's keep this code, but at least we should have an end-to-end test that inserts and selects slice/map from a real table...

PR #1934, opened 2021-12-01
add receiver method ToURL to *Options

dillonstreator:
discussion reference https://github.com/go-pg/pg/discussions/1933

### DIFF
--- a/options.go
+++ b/options.go
@@ -256,6 +256,53 @@ func ParseURL(sURL string) (*Options, error) {
 	return options, nil
 }
 
+func (opts *Options) ToURL() string {
+	dsn := "postgres://"
+
+	if len(opts.User) > 0 {
+		dsn += opts.User
+
+		if len(opts.Password) > 0 {
+			dsn += ":" + opts.Password
+		}
+
+		dsn += "@"
+	}
+
+	if len(opts.Addr) > 0 {
+		dsn += opts.Addr
+	} else {
+		dsn += "localhost:5432"
+	}
+
+	dsn += "/" + opts.Database
+
+	values := url.Values{}
+
+	if opts.DialTimeout > 0 {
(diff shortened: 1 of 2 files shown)


options.go
```diff
@@ -256,6 +256,53 @@ func ParseURL(sURL string) (*Options, error) {
...
+		values.Add("application_name", opts.ApplicationName)
+	}
+
+	if opts.TLSConfig == nil {
+		values.Add("sslmode", "disable")
+	} else if opts.TLSConfig.InsecureSkipVerify {
+		values.Add("sslmode", "allow")
+	} else if !opts.TLSConfig.InsecureSkipVerify {
+		values.Add("sslmode", "verify-ca")
+	}
```
elliotcourant, line 296: An else at the end that also includes `sslmode=disable` might not be a bad idea?

options_test.go
```diff
@@ -261,3 +261,29 @@ func TestParseURL(t *testing.T) {
 		})
 	}
 }
+
+var hello struct {
+	Test string
+}
+
+func TestOptions_ToURL(t *testing.T) {
```
elliotcourant, line 267: Another good test case might be to try to parse the URL that is output from ToURL to make sure that it is still a valid input.

elliotcourant (changes requested):

options_test.go
dillonstreator: yeah that makes sense to me.. good call

options.go
dillonstreator: @elliotcourant an `else` branch wouldn't be possible to hit seeing that if `TLSConfig` is not `null`, given the zero value of `bool`, either of the `else if` cases would be guaranteed to execute.

elliotcourant (approved):

PR #1946, opened 2022-03-10
Fix hook

oligot:
With the current version, I have those errors when trying to use the
hook:
```
cannot use pgdebug.DebugHook{...} (type pgdebug.DebugHook) as type pg.QueryHook in argument to db.baseDB.AddQueryHook:
pgdebug.DebugHook does not implement pg.QueryHook (BeforeQuery method has pointer receiver)
```

```
cannot use pgotel.TracingHook{} (type pgotel.TracingHook) as type pg.QueryHook in argument to db.baseDB.AddQueryHook:
pgotel.TracingHook does not implement pg.QueryHook (AfterQuery method has pointer receiver)
```

### DIFF
--- a/extra/pgdebug/pgdebug.go
+++ b/extra/pgdebug/pgdebug.go
@@ -23,7 +23,7 @@ func NewDebugHook() *DebugHook {
 
 var _ pg.QueryHook = (*DebugHook)(nil)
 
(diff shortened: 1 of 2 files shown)


oligot:
My bad, I wasn't using the NewxxxHook (NewDebugHook and NewTracingHook) methods.

PR #1947, opened 2022-03-16
Add shortcut WhereInOr

swad2007:
A shortcut is proposed for implementing a query to the database, for example, SELECT * WHERE column 1 in the slice or column 2 in the slice

### DIFF
--- a/orm/query.go
+++ b/orm/query.go
@@ -570,6 +570,11 @@ func (q *Query) WhereIn(where string, slice interface{}) *Query {
 	return q.Where(where, types.In(slice))
 }
 
+// WhereInOr is a shortcut for WhereOr and pg.In.
+func (q *Query) WhereInOr(where string, slice interface{}) *Query {
+	return q.WhereOr(where, types.In(slice))
+}
+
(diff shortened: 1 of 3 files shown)


version.go
```diff
@@ -2,5 +2,5 @@ package pg
 
 // Version is the current release version.
 func Version() string {
-	return "10.10.6"
```
elliotcourant: I would not increment this please.

orm/query.go
```diff
@@ -570,6 +570,11 @@ func (q *Query) WhereIn(where string, slice interface{}) *Query {
 	return q.Where(where, types.In(slice))
 }
 
+// WhereInOr is a shortcut for WhereOr and pg.In.
+func (q *Query) WhereInOr(where string, slice interface{}) *Query {
+	return q.WhereOr(where, types.In(slice))
+}
```
elliotcourant, line 576: Would you be able to add a test proving the query is built how you intend it to be built with this function. Like in `orm/select_nil_test.go`

elliotcourant (changes requested):

elliotcourant (approved):

PR #1948, opened 2022-04-13
feat(orm): add materialization to SELECT CTE statements

maximerety:
For emergency situations where we would want to force CTE materialization:

```go
func (q *Query) With(name string, subq *Query) *Query                // WITH "name" AS (...)
func (q *Query) WithMaterialized(name string, subq *Query) *Query    // WITH "name" AS MATERIALIZED (...)
func (q *Query) WithNotMaterialized(name string, subq *Query) *Query // WITH "name" AS NOT MATERIALIZED (...)
```

### DIFF
--- a/orm/query.go
+++ b/orm/query.go
@@ -36,9 +36,18 @@ const (
 	allWithDeletedFlag
 )
 
+type materializationMode uint8
+
+const (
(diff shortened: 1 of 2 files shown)


maximerety:
Merge request created upstream by mistake. Sorry about the noise 🙇

PR #1951, opened 2022-07-25
fix buff size comparison & proper error message

tufanbarisyildirim:
When scanning a byte column `ReadBytes` expect to have exact size of a `[]byte` We should be able to read bytes into a buffer that is bigger than our data size in this column. This cause errors when you have a fixed field size in code but variable size in database. The error message was also misleading like it was checking if we have enough space but it actually expect to have exact size. So, changed the comparison to accept bigger buffers and to properly print the source & destination buffer size.

### DIFF
--- a/types/scan.go
+++ b/types/scan.go
@@ -107,8 +107,8 @@ func ReadBytes(rd Reader, b []byte) error {
 	}
 	tmp = tmp[2:] // Trim off "\\x".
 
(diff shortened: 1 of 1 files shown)


elliotcourant:
Is there a way that this behavior could be proven with a unit test, even if that test is just for the ReadBytes function itself?

PR #1964, opened 2022-10-08
Release v10.10.7 (release.sh)

elliotcourant:


elliotcourant:
```
--- FAIL: ExampleDB_Model_createTable (0.02s)
got:
[{id bigint} {model1_id bigint} {name text}]
want:
[{id bigint} {name text} {model1_id bigint}]
```

Got this failure which is just a fluke.

PR #1981, opened 2023-06-20
Update err for single conn pool when context.Done() return first

xin-tsla:
Hi All,
The issue this PR tries to solve is that we observed an NPE in `OnConnect` method 
 https://github.com/go-pg/pg/blob/f7e1c98fed3044ae4e195a0e063f69185fb1885b/base.go#L119
when we call `conn.Exec()` twice inside the `OnConnect` method. and the second call (`conn.Exec()`) may get an `nil` conn object and fails to call `conn.Exec()`[panic].

The data flow diagram depicts how the conn is `nil` in some case:
ctxcancelledsingleconnpool drawio

If `ctx` gets cancelled and returns before query finishes (Yellow line in the diagram), the `conn` in `SingleConnPool` will call `Remove` method to set the conn to `nil`:
https://github.com/go-pg/pg/blob/f7e1c98fed3044ae4e195a0e063f69185fb1885b/internal/pool/pool_single.go#L38
and as well to set an error with reason, which is `nil` since it is due to `ctx` gets cancelled without an error:
https://github.com/go-pg/pg/blob/f7e1c98fed3044ae4e195a0e063f69185fb1885b/internal/pool/pool_single.go#L39

When `ctx` is cancelled:
https://github.com/go-pg/pg/blob/f7e1c98fed3044ae4e195a0e063f69185fb1885b/base.go#L153
and send `struct{}{}` to fnDone:
https://github.com/go-pg/pg/blob/f7e1c98fed3044ae4e195a0e063f69185fb1885b/base.go#L159

### DIFF
--- a/internal/pool/pool_single.go
+++ b/internal/pool/pool_single.go
@@ -37,6 +37,9 @@ func (p *SingleConnPool) Put(ctx context.Context, cn *Conn) {}
 func (p *SingleConnPool) Remove(ctx context.Context, cn *Conn, reason error) {
 	p.cn = nil
 	p.stickyErr = reason
+	if reason == nil && ctx != nil {
+		p.stickyErr = ctx.Err()
+	}
 }
 
 func (p *SingleConnPool) Close() error {
--- a/internal/pool/pool_single.go
+++ b/internal/pool/pool_single.go
@@ -37,6 +37,8 @@ func (p *SingleConnPool) Put(ctx context.Context, cn *Conn) {}
 func (p *SingleConnPool) Remove(ctx context.Context, cn *Conn, reason error) {
 	p.cn = nil
 	p.stickyErr = reason
+	// If ctx is cancelled without a reason(error) value,
+	// then the ctx.Error is used as the reason for why the p.cn is assigned nil.
 	if reason == nil && ctx != nil {
 		p.stickyErr = ctx.Err()
 	}
--- a/internal/pool/pool_single_test.go
+++ b/internal/pool/pool_single_test.go
@@ -11,12 +11,12 @@ import (
 var _ = Describe("SingleConnPool", func() {
 	It("remove a conn due to context is cancelled", func() {
 		p := pool.NewSingleConnPool(nil, &pool.Conn{})
-		ctx, cl := context.WithCancel(context.TODO())
+		ctx, cancel := context.WithCancel(context.TODO())
 		cn, err := p.Get(nil)
 		Expect(err).To(BeNil())
 		Expect(cn).ToNot(BeNil())
 
-		cl()
+		cancel()
 		p.Remove(ctx, cn, nil)
 		cn, err = p.Get(nil)
 		Expect(cn).To(BeNil())
--- a/internal/pool/pool_single_test.go
+++ b/internal/pool/pool_single_test.go
@@ -1,0 +1,31 @@
+package pool_test
+
+import (
+	"context"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/go-pg/pg/v10/internal/pool"
+)
+
+var _ = Describe("SingleConnPool", func() {
+	var p *pool.SingleConnPool
+
+	BeforeEach(func() {
+		p = pool.NewSingleConnPool(nil, &pool.Conn{})
+	})
+
+	It("remove a conn due to context is cancelled", func() {
+		ctx, cl := context.WithCancel(context.TODO())
+		cn, err := p.Get(nil)
+		Expect(err).To(BeNil())
+		Expect(cn).ToNot(BeNil())
+
+		cl()
+		p.Remove(ctx, cn, nil)
+		cn, err = p.Get(nil)
+		Expect(cn).To(BeNil())
+		Expect(err).ToNot(BeNil())
+	})
+
+})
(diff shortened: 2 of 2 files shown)


internal/pool/pool_single_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/go-pg/pg/v10/internal/pool"
+)
+
+var _ = Describe("SingleConnPool", func() {
+	var p *pool.SingleConnPool
+
+	BeforeEach(func() {
+		p = pool.NewSingleConnPool(nil, &pool.Conn{})
+	})
```
elliotcourant: Would this need to be a before each and a variable in this scope? Or could the pool just get moved into the test itself since there is only one?

internal/pool/pool_single_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	It("remove a conn due to context is cancelled", func() {
+		ctx, cl := context.WithCancel(context.TODO())
```
elliotcourant: I'd rename `cl` to `cancel` to make it clear here.

internal/pool/pool_single_test.go
```diff
@@ -0,0 +1,31 @@
...
+		Expect(err).To(BeNil())
+		Expect(cn).ToNot(BeNil())
+
+		cl()
+		p.Remove(ctx, cn, nil)
+		cn, err = p.Get(nil)
+		Expect(cn).To(BeNil())
+		Expect(err).ToNot(BeNil())
+	})
+
```
elliotcourant: nit pick, no need for this extra newline

internal/pool/pool_single.go
```diff
@@ -37,6 +37,9 @@ func (p *SingleConnPool) Put(ctx context.Context, cn *Conn) {}
 func (p *SingleConnPool) Remove(ctx context.Context, cn *Conn, reason error) {
 	p.cn = nil
 	p.stickyErr = reason
+	if reason == nil && ctx != nil {
+		p.stickyErr = ctx.Err()
+	}
```
elliotcourant, line 44: Just to make sure I'm understanding it correctly.

When a query/connection is canceled, we will not receive an error as a reason. Hence the nil check here. So when we are removing it from the connection pool and we haven't been given a reason; check to see if there is an error on the context itself, which will have either a cancel error or some kind of deadline exceed.

Is that correct?

---

If so could that be left as a hand wavy comment here? The diagram in the PR is excellent for the flow of events, but hard to back reference that if changes ever need to be made in the future. So having comments is always nice.

elliotcourant (changes requested):
Overall this looks good to me, I have a things I'd want to just tweak/get clarification on. But I'm happy to merge this once thats ready!

internal/pool/pool_single_test.go
xin-tsla: Good point! It is more clean, addressed: 9653a2859e14eb96aa859f06a4a2c513230edabc

internal/pool/pool_single_test.go
xin-tsla: Yes, renamed: ec32de71bf6b8cf79af72db3011ffa4572c753c4

internal/pool/pool_single_test.go
xin-tsla: Addressed along with a previous commit https://github.com/go-pg/pg/commit/9653a2859e14eb96aa859f06a4a2c513230edabc

internal/pool/pool_single.go
xin-tsla: @elliotcourant , yes, that is correct.
The `reason` (`err`) is nil, since there is no error returns from the query, instead ctx.Done() returns with either `context cancelled` or `deadline exceed`.
Since `err` is nil, so it did not have an error value in line 76.
https://github.com/go-pg/pg/blob/f7e1c98fed3044ae4e195a0e063f69185fb1885b/base.go#L76
And in line 80, then `nil` conn object checks its `Inited` field and causes an NPE:

https://github.com/go-pg/pg/blob/f7e1c98fed3044ae4e195a0e063f69185fb1885b/base.go#LL80C1

image

Good point! I will add a comment for explaining the logic.

xin-tsla: Added a comment: 2755f6daad1fb5d0eaaa204c48afc7e4ba63a1c4

xin-tsla:
Hi @elliotcourant , wondering when you have time, could you take a second look?

elliotcourant (approved):

elliotcourant:
Thank you very much for the fix @xin-tsla !

xin-tsla:
@elliotcourant , happy to work on the fix! And much appreciate your review and time! :)

xin-tsla:
@elliotcourant , when you have time, another question, could you create a new release version?

elliotcourant:
I can tag it for release this week, gonna try to loop in some other changes with that release as well. Mostly dependency bumps.

xin-tsla:
@elliotcourant , got it. Thank you so much for tagging the release! :)

xin-tsla:
@elliotcourant , by chance when you have time, could you tag a new release version?

elliotcourant:
@xin-tsla Sorry for the delay! It is now tagged!

xin-tsla:
@elliotcourant , no worries! Thanks for tagging it! Much appreciate your time!

Issue #1979, opened 2023-05-24
Scan JSON value get panic


nirsht:
https://github.com/go-pg/pg/blob/f7e1c98fed3044ae4e195a0e063f69185fb1885b/types/scan_value.go#LL254C2-L254C2
If the JSON value , this line get panic.

Here is the stack trace

```
/usr/local/go/src/runtime/panic.go:838 +0x207
reflect.flag.mustBeAssignableSlow(0x0?)
	/usr/local/go/src/reflect/value.go:262 +0x85
reflect.flag.mustBeAssignable(...)
	/usr/local/go/src/reflect/value.go:249
reflect.Value.Set({0x12609e0?, 0xc005366db0?, 0x12609e0?}, {0x12609e0?, 0xc0031bc508?, 0x1f17ca0?})
	/usr/local/go/src/reflect/value.go:2084 +0x6a
github.com/go-pg/pg/v10/types.scanJSONValue({0x12609e0?, 0xc005366db0?, 0xc00b8f23b8?}, {0x1618b08?, 0xc01080a1e0}, 0x791)
	/go/pkg/mod/github.com/go-pg/pg/v10@v10.10.6/types/scan_value.go:254 +0x7a
github.com/go-pg/pg/v10/types.ScanValue({0x12609e0?, 0xc005366db0?, 0x13a0fc0?}, {0x1618b08, 0xc01080a1e0}, 0xc003c08340?)
	/go/pkg/mod/github.com/go-pg/pg/v10@v10.10.6/types/scan_value.go:177 +0x13a
github.com/go-pg/pg/v10/types.scanIfaceValue({0x1241bc0?, 0xc031f801b8?, 0x0?}, {0x1618b08, 0xc01080a1e0}, 0x4?)
	/go/pkg/mod/github.com/go-pg/pg/v10@v10.10.6/types/scan_value.go:167 +0xb2
github.com/go-pg/pg/v10/orm.(*Field).ScanValue(0xc002f5ec80, {0x132e1a0?, 0xc031f80000?, 0x4?}, {0x1618b08, 0xc01080a1e0}, 0x791)
	/go/pkg/mod/github.com/go-pg/pg/v10@v10.10.6/orm/field.go:125 +0xd4
github.com/go-pg/pg/v10/orm.(*structTableModel).scanColumn(0xc003b149c0, {0x3c01?, 0xc0?, {0xc02504c200?, 0xb?}}, {0x1618b08, 0xc01080a1e0}, 0x791)
[...]

PR #1984, opened 2023-08-11
new instance of the struct type for scanjsonvalue #1979

Slitherings:
First pull request, wasn't able to test or make sure but this code looks like a fix towards the Scan JSON value get panic #1979
issue (not sure about the "return nil" code on 269) but aside from that hopefully this can be a step towards fixing this issue, but I saw this as a good time to push a potential fix since pg is in the process of migrating to Bun.

### DIFF
--- a/types/scan_value.go
+++ b/types/scan_value.go
@@ -249,16 +249,24 @@ func scanStringValue(v reflect.Value, rd Reader, n int) error {
 }
 
 func scanJSONValue(v reflect.Value, rd Reader, n int) error {
-	// Zero value so it works with SelectOrInsert.
-	// TODO: better handle slices
-	v.Set(reflect.New(v.Type()).Elem())
-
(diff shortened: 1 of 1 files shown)


elliotcourant:
I kind of dropped the ball on that issue and never looped back to it sorry. I can take a closer look at this code locally soon and see if I can come up with a test or point you in a direction for a test.

Slitherings:
> I kind of dropped the ball on that issue and never looped back to it sorry. I can take a closer look at this code locally soon and see if I can come up with a test or point you in a direction for a test.

All good, keep me posted just trying to help!

elliotcourant:
@Slitherings is there a unit test to support this change?

elliotcourant:
Sorry pushing all the wrong buttons, I need to figure out how to get this tested. Ideally we should have a unit test that would have failed before this patch but passes with it.

PR #1988, opened 2023-11-06
Optimize memory allocation in appendRune function

### LINKED ISSUE #1987: Potential Memory Leak in appendRune Function
## Expected Behavior

The appendRune function should handle memory allocation efficiently, especially when dealing with large strings.

## Current Behavior

The current implementation of appendRune function in go-pg library seems to be causing a memory leak. It creates a new byte slice every time the capacity of the current slice is insufficient to store the new rune. This can lead to a large number of memory allocations, especially when the function is called with large strings.

## Possible Solution

A possible solution would be to modify the appendRune function to reuse the already allocated memory as much as possible, instead of always allocating new memory. This can be achieved by doubling the capacity of the existing slice when the current capacity is insufficient.

## Steps to Reproduce

1. Call the appendRune function with a large string.
2. Monitor the memory usage.
3. Notice the increase in memory allocation.

## Context (Environment)

This issue has been identified while profiling a Go application that uses the go-pg library. The memory leak can potentially affect the performance and stability of applications that use this library.

## Detailed Description

The proposed change involves modifying the appendRune function in the go-pg library to handle memory allocation more efficiently. This should help mitigate the memory leak issue and improve the performance of applications using this library.

## Possible Implementation

Here is a possible implementation of the proposed change:

```go
func appendRune(b []byte, r rune) []byte {
    if r < utf8.RuneSelf {
        return append(b, byte(r))
    }
    l := len(b)
    if cap(b)-l < utf8.UTFMax {
        nb := make([]byte, l, 2*l+utf8.UTFMax)
        copy(nb, b)
        b = nb
    }
    n := utf8.EncodeRune(b[l:l+utf8.UTFMax], r)
    return b[:l+n]
}
```
In this version of the function, instead of always creating a new byte slice when the capacity is insufficient, we double the capacity of the existing slice. This should reduce the amount of memory allocated and, therefore, the memory leak.

mateusveloso:
This pull request addresses a potential memory leak in the appendRune function of the go-pg library. The current implementation of the function creates a new byte slice every time the capacity of the current slice is insufficient to store the new rune. This can lead to a large number of memory allocations, especially when the function is called with large strings.

The proposed change modifies the appendRune function to reuse the already allocated memory as much as possible, instead of always allocating new memory. This is achieved by doubling the capacity of the existing slice when the current capacity is insufficient.

This change should help mitigate the memory leak issue and improve the performance of applications using the go-pg library.

Here is the modified version of the appendRune function:

```go
func appendRune(b []byte, r rune) []byte {
    if r < utf8.RuneSelf {
        return append(b, byte(r))
    }
    l := len(b)
    if cap(b)-l < utf8.UTFMax {
        nb := make([]byte, l, 2*l+utf8.UTFMax)
        copy(nb, b)
        b = nb
    }
    n := utf8.EncodeRune(b[l:l+utf8.UTFMax], r)
    return b[:l+n]
}
```
This PR resolves [#1987](https://github.com/go-pg/pg/issues/1987)

### DIFF
--- a/types/append.go
+++ b/types/append.go
@@ -160,7 +160,9 @@ func appendRune(b []byte, r rune) []byte {
 	}
 	l := len(b)
 	if cap(b)-l < utf8.UTFMax {
-		b = append(b, make([]byte, utf8.UTFMax)...)
+		nb := make([]byte, l, 2*l+utf8.UTFMax)
+		copy(nb, b)
+		b = nb
 	}
 	n := utf8.EncodeRune(b[l:l+utf8.UTFMax], r)
 	return b[:l+n]
--- a/types/append_test.go
+++ b/types/append_test.go
@@ -1,0 +1,56 @@
+package types
+
+import (
+	"testing"
+	"unicode/utf8"
+)
+
+func BenchmarkAppendRuneOld(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		s := make([]byte, 0, 1024)
+		b.StartTimer()
+
+		for j := 0; j < 1000000; j++ {
+			s = appendRuneOld(s, '世')
+		}
+	}
+}
+
+func BenchmarkAppendRuneNew(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		s := make([]byte, 0, 1024)
+		b.StartTimer()
+
+		for j := 0; j < 1000000; j++ {
+			s = appendRuneNew(s, '世')
+		}
+	}
+}
+
+func appendRuneOld(b []byte, r rune) []byte {
+	if r < utf8.RuneSelf {
+		return append(b, byte(r))
+	}
+	l := len(b)
+	if cap(b)-l < utf8.UTFMax {
+		b = append(b, make([]byte, utf8.UTFMax)...)
+	}
+	n := utf8.EncodeRune(b[l:l+utf8.UTFMax], r)
+	return b[:l+n]
+}
+
+func appendRuneNew(b []byte, r rune) []byte {
+	if r < utf8.RuneSelf {
+		return append(b, byte(r))
+	}
+	l := len(b)
+	if cap(b)-l < utf8.UTFMax {
+		nb := make([]byte, l, 2*l+utf8.UTFMax)
+		copy(nb, b)
+		b = nb
+	}
+	n := utf8.EncodeRune(b[l:l+utf8.UTFMax], r)
+	return b[:l+n]
+}
--- a/types/append_test.go
+++ b/types/append_test.go
@@ -2,20 +2,7 @@ package types
 
 import (
 	"testing"
-	"unicode/utf8"
 )
-
-func BenchmarkAppendRuneOld(b *testing.B) {
-	for i := 0; i < b.N; i++ {
-		b.StopTimer()
-		s := make([]byte, 0, 1024)
-		b.StartTimer()
-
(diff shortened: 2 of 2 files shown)


elliotcourant:
From my understanding of the append rune function, this would only improve memory allocation when you are frequently appending runes that are multiple bytes by pre-allocating some space for those runes. But if you are infrequently doing this then you would be over-allocating.

The current code allocates exactly what is needed but would do it far more frequently (and potentially inefficiently if you are frequently using multi-byte runes), the new code solves this but I don't interpret it as a fix but more as a tradeoff for a specific use case?

Can you provide more details or tests that demonstrate the memory behavior that you were observing with the old code and how those are resolved with the new approach?

elliotcourant:
I would be particularly interested in a unit test for this function that could be run on the primary branch and this one with a memprofile so that we could compare.

mateusveloso:
Absolutely, I understand the importance of providing empirical evidence to support this change. I have set up a unit test for the `appendRune` function that can be run with a memory profiler on both the primary branch and this one. This allows us to compare the memory usage of both implementations.

The benchmark test was designed to stress the `appendRune` function by repeatedly appending a multi-byte rune to a byte slice. The test was run with a large number of iterations to simulate a high-load scenario and make any memory leak more evident.

Here are the results of the benchmark test:

- Old implementation: `BenchmarkAppendRuneOld-16 136 8829282 ns/op 16790812 B/op 30 allocs/op`
- New implementation: `BenchmarkAppendRuneNew-16 168 7076402 ns/op 8456451 B/op 12 allocs/op`

As you can see, the new implementation of the `appendRune` function allocates significantly less memory per operation (8.45 MB vs 16.79 MB) and makes fewer distinct allocations (12 vs 30). This suggests that the new implementation is more efficient in terms of memory allocation, which can help mitigate the memory leak issue.

I believe these results demonstrate the effectiveness of the proposed change in improving the performance of the `appendRune` function. I hope this addresses your concerns, and I'm happy to provide any further information or make any additional changes as needed.

@elliotcourant

elliotcourant:
@MateusVeloso outstanding! Thank you very much for putting that together, I'm gonna play around with it locally a bit tonight with non-multibyte runes and look at a difference in allocation there as well. But so far this looks good to me!

Out of curiosity, was this discovered due to a use case in your app with primarily multi-byte runes from user input? Names, descriptions, metadata etc?

mateusveloso:
@elliotcourant I discovered this issue while working on my application, which handles a significant amount of user input, including multi-byte runes in names, descriptions, metadata, etc.

I noticed an unusual memory usage pattern that led me to suspect a potential memory leak. To investigate further, I utilized the go tool pprof.

Upon analysis, I found that the `appendRune` function was a significant contributor to the memory usage. Given the frequency and volume of data my application processes, optimizing this function seemed like a promising avenue for improving overall performance and efficiency.

I'm glad to hear that the changes look good to you so far. I'm confident that they will prove beneficial not only for my application but also for other developers using the go-pg library. I look forward to hearing about your experiences testing it with non-multibyte runes.

elliotcourant:
@MateusVeloso I'm fine to merge this, would you mind removing the benchmark for the "old" version and the old function? But leave the benchmark around the new one in?

mateusveloso:
@elliotcourant  
Absolutely, I appreciate the feedback! I'll go ahead and remove the benchmark for the old version and the old function as suggested.

mateusveloso:
@elliotcourant I've removed the old benchmark

elliotcourant (approved):

PR #1996, opened 2024-01-23
Add password environment variable for customizing running tests

DarrylWong:
In #1429, environment variables were added to read in postgres connection details for customizing tests. This change does the same but with a password variable.

It also removes the explicit user and password values from pgOptions, as Options init() will set it as postgres anyway if not specified. This also enables env variables to be used instead for those values.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -14,12 +14,11 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


elliotcourant:
Seems reasonable to me, thank you @DarrylWong

PR #1997, opened 2024-02-06
Add WhereOrg and WhereOrgAndPK methods

bennichols:


elliotcourant:
I'm probably not going to merge this, these functions are very specific to (what I assume is) your use case regarding go-pg. It introduces a dependency for only a single function call, doesn't include tests, and is not safe with joins (if two tables have `organization_id` then this breaks the query).

Thank you for your contribution, but I think if this is something you need from go-pg, I would recommend forking the library and using the fork instead. This library does not receive frequent updates anymore and most of the changes are bug fixes at this point. So you will have little risk of falling behind on upstream functionality.

bennichols:
My apologies, I didn't mean to make a PR for this to the main repo, just to the fork!

PR #2003, opened 2024-04-28
generics model support

cubatic45:
generics model support


orm/table.go
```diff
@@ -95,9 +95,14 @@ type Table struct {
...
+	name := t.Type.Name()
+	index := strings.Index(name, "[")
+	if index > 0 {
+		name = name[:index]
+	}
 	t.zeroStruct = reflect.New(t.Type).Elem()
-	t.TypeName = internal.ToExported(t.Type.Name())
-	t.ModelName = internal.Underscore(t.Type.Name())
+	t.TypeName = internal.ToExported(name)
+	t.ModelName = internal.Underscore(name)
```
elliotcourant, line 105: Wouldn't an easier way to do this be to just define the name directly on the thing?

Like this:

```go
type File struct {
	tableName string `pg:"files"`

	...
```

This way even if its a generic, the table name stays the same? With how you have it written, you're throwing out the generic argument entirely, so it would never be considered for the table name.

cubatic45: Directly defining the name works in some models without nested structure, but in some cases, it does not work.

Sometimes, we forget to use an alias.
example: https://go.dev/play/p/VJpB_c3cLkt

elliotcourant: Oh interesting, that is definitely a bug in my eyes then, the way table names are built should be consistent. Let me mess around a bit here. 

Thank you very much for the code reproduction!

elliotcourant: I see now what is happening, the way we generate the alias is different than how we derive the actual table name. Thus why it is only affecting select queries oddly.

I think there is a more elegant solution to this, but let me mess around some more.

cubatic45: Have you found a better solution?

PR #2006, opened 2024-07-02
add AppendColumn method

bohdand-weka:


orm/query.go
```diff
@@ -482,60 +506,60 @@ func (q *Query) WhereOr(condition string, params ...interface{}) *Query {
...
-//    q.Where("TRUE").
-//    	WhereGroup(func(q *pg.Query) (*pg.Query, error) {
-//    		q = q.WhereOr("FALSE").WhereOr("TRUE").
-//    		return q, nil
-//    	})
+//	q.Where("TRUE").
+//		WhereGroup(func(q *pg.Query) (*pg.Query, error) {
+//			q = q.WhereOr("FALSE").WhereOr("TRUE").
+//			return q, nil
+//		})
```
elliotcourant: Could you undo all these whitespace changes please?

elliotcourant (changes requested):
I'm curious what the use case for this is, as the `AppendColumn` method itself only adds columns that are present on the actual model you're working with. If you are joining between two tables, it would only let you append columns from the base table (whatever was passed to `Model(...)`. Which all of the columns from that model would already be present by default?

Would you be able to provide an example use case for this as well as tests?

bohdand-weka:
> I'm curious what the use case for this is, as the `AppendColumn` method itself only adds columns that are present on the actual model you're working with. If you are joining between two tables, it would only let you append columns from the base table (whatever was passed to `Model(...)`. Which all of the columns from that model would already be present by default?
> 
> Would you be able to provide an example use case for this as well as tests?

The use case is simple. 

Imagine we have dynamic columns selection. By default `Column()` will override all model columns when called first time.

So in our scenario we don't want to override default columns if `Column` wasn't called with `Alias.*`, instead we want to just append one more column from our calculation, like aggregation or whatever.
If in our dynamic query there was already `Column` called - `AppendColumn` won't touch them, so it does exactly how it's called - appends column.

And it will work for joins as well (as like `Column` does), why not?

orm/query.go
bohdand-weka: sure, looks like it came from vscode

PR #2007, opened 2024-07-04
Update pgutil_test.go

AndersWoodruff:
Fix import of pgutil to prevent error. As of now it causes the following error:

```
       [gopkg.in/pg.v3](http://gopkg.in/pg.v3) imports

	[gopkg.in/pg.v3/pgutil](http://gopkg.in/pg.v3/pgutil) tested by

	[gopkg.in/pg.v3/pgutil.test](http://gopkg.in/pg.v3/pgutil.test) imports

	[github.com/go-pg/pg/pgutil](http://github.com/go-pg/pg/pgutil): module [github.com/go-pg/pg@latest](http://github.com/go-pg/pg@latest) found (v8.0.7+incompatible), but does not contain package [github.com/go-pg/pg/pgutil](http://github.com/go-pg/pg/pgutil)

```

### DIFF
--- a/pgutil/pgutil_test.go
+++ b/pgutil/pgutil_test.go
@@ -3,7 +3,7 @@ package pgutil_test
 import (
 	"testing"
 
-	"github.com/go-pg/pg/pgutil"
+	"gopkg.in/pg.v3/pgutil"
 )
 
 func TestUnderscore(t *testing.T) {


elliotcourant:
This v3 branch is almost 10 years old, and predates go modules entirely. While it may have gone by `gopkg.in/pg.v3` then, it does not anymore. I would really just recommend upgrading to the v10 version which is still somewhat maintained.

AndersWoodruff:
Sure, thanks for the response. I agree that upgrading makes sense, but in the mean time I think this fix should allow the old version to run successfully, which at the moment I believe it does not.

Unfortunately upgrading would require many changes to our codebase, as we use functions that are not in the newer versions.

elliotcourant:
In that case I would recommend updating your code to work off of a fork. I'm not going to merge this to a version of the code I've never even touched before, sorry.

PR #2008, opened 2024-07-08
fix: allow to exclude more complex column definition

bohdand-weka:
Right now ExcludeColumn works only for simple cases, for example 
```
q.Column("alias.*").Exclude("*")
```
will not exclude this column.

By using regexp for parsing alias and column in exclude column method it will be possible to exclude more complex selectors, also added support for `SafeQueryAppender`

### DIFF
--- a/orm/query.go
+++ b/orm/query.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


elliotcourant:
This is far too complicated to merge to this project at this point. go-pg is basically maintenance mode only, and while Vladimir let me take over some of this stuff as I use go-pg heavily for work. At this point I might as well be the only maintainer, and this PR is too much for me to want to take on if something doesn't go well in the future. It's too large of a feature to add.


I'm sorry, I would really recommend forking go-pg and adding these features there and periodically pulling from upstream if there are any fixes.

bohdand-weka:
@elliotcourant we already made fork for us , this PR is just to share it with upstream (might be useful for others)

PR #2014, opened 2024-12-12
Make reader and writer buffers configurable

tony2001:
Replace global sync pools with per-DB pools and make reader and writer buffers configurable. 
Default buffer sizes: 1Mb for reader buffer, 64Kb for writer buffer.

Context:
DB connections are pooled, usually there are not much of them available, so this is a resource used by goroutines, which have to wait for their turn in order to get a connection and use it before returning it back to the pool.

Before getting into the waiting line every goroutine allocates a read and a write buffer from the sync.Pool of buffers.
Currently hardcoded reader buffer size is 1Mb, so when 1000 goroutines wait in the queue, you get 1000Mb of buffers pre-allocated. So when an application unexpectedly gets a spike of traffic and all the database connections are being used, we don't get request timeouts as one could be expecting, the application is being OOM-killed instead.

The patch addresses this issue by trading some allocations and (probably, though my benchmarks don't really show it) latency for the ability to serve more simultaneous connections.

### DIFF
--- a/base.go
+++ b/base.go
@@ -227,8 +227,8 @@ func (db *baseDB) ExecContext(c context.Context, query interface{}, params ...in
 }
 
 func (db *baseDB) exec(ctx context.Context, query interface{}, params ...interface{}) (Result, error) {
-	wb := pool.GetWriteBuffer()
-	defer pool.PutWriteBuffer(wb)
+	wb := db.pool.GetWriteBuffer()
+	defer db.pool.PutWriteBuffer(wb)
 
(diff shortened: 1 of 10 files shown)


elliotcourant:
At a glance this looks fine to me, let me pull it down locally this afternoon and poke around a bit with it.

elliotcourant (approved):
I think this makes sense, it also cleans up a global variable which I think is good. But I'm wondering if now initiating the pool per DB instance will incur any additional cost, like when running tests where a DB instance might be created per test.

Overall the tradeoff though I think is worth it even if that does have a negative effect (which I'm not sure it even would).

PR #2018, opened 2025-03-28
fix(orm): create sql.NullInt16 column as smallint, sql.NullInt32 as integer

lightsgoout:
Currently, sql.NullInt16 and sql.NullInt32 are created as JSONB columns. Fix that.


elliotcourant:
I'll want to play with this a bit before merging but if you want to fix it in the mean time; you can add the `type:"smallint"` or `type:"integer"` go tag to those fields on your struct and that should correct the issue in the mean time.

PR #2039, opened 2025-11-26
fix(deps): Fixing x/crypto/ssh security warnings

elliotcourant:


elliotcourant:
This is a bit of a hack however, instead of bumping all of the (basically stable) dependencies that go-pg is using. It seems better to just bump the transitive dependency that is the `/x/...` packages as they don't _usually_ have breaking changes and as long as tests pass should continue to work just fine.

This way the security alert is properly resolved and any others that came with any of the `/x/...` dependencies.

